Issue #22, opened 2014-10-07
Review each guide and improve the content and structure


mattab:
The goal of this project is a medium term goal that you review each page of the Developer website. The goal is to make the documentation from OK quality to good or even very good (!) when possible. 

Project description:

 * Docs to review: https://github.com/piwik/developer-documentation/tree/master/docs
 * There are about 40 pages some of them maybe un-necessary, some can be improved, some could be maybe merged or split. 
 * Good resource: http://jacobian.org/writing/great-documentation/

See also the related issue in [piwik/piwik](https://github.com/piwik/piwik/issues/4456#ticket)

Assigned to at least @mnapoli and @mattab who will review content of each page

- Integrate
  - [x] #61 The Javascript Tracking Client
  - [x] #41 Tracking API Clients
  - [x] 60b521d2e6424654a6613537754fc76609efb739 All About Tracking
  - [x] #65 Content Tracking
  - [x] #66 Reporting API Tutorial: Get your top 10 keywords
  - [x] #66 Querying the Reporting API
- Develop
  - [x] #25 Getting Started Part I
  - [x] #26 Getting Started Part II
  - [x] 97ad26c390b Getting Started Part III
  - [x] #49 Distributing Your Plugin
  - [x] #32 How Piwik Works
  - [x] #64 All About Analytics Data
  - [x] #60 Piwik's Extensibility Points
  - [x] b94d9c6b8b5f MVC in Piwik
  - [x] #48 Pages
  - [x] #48 Menus
  - [x] #48 Widgets
  - [x] #69 JavaScript and CSS
  - [x] #59 Visualizing Report Data
  - [x] #43 Piwik INI Configuration (moved as sub-article of Piwik Configuration)
  - [x] #43 Piwik Configuration (extracted "Plugin Settings" into sub-article)
  - [x] #43 Plugin Settings
  - [x] #56 Persistence & the MySQL Backend
  - [x] #36 Security best practices
  - [x] #48 Permissions
  - [x] #37 Internationalization
  - [x] #38 Tests
  - [x] #63 Piwik's Reporting API
  - [x] #50 Scheduled tasks
  - [x] #52 Piwik on the Command Line
  - [x] #53 Contributing to Piwik Core
  - [x] #30 Core Team Workflow
  - [x] #70 Theming

mnapoli:
Once #70 is merged this task is finished!

mnapoli:
Done!

mattab:
[...]

PR #59, opened 2014-12-21
Reviewed the "Display Reports" guide

mnapoli:
#22


mattab:
:+1: http://developer.piwik.org/guides/visualizing-report-data looks good!

PR #66, opened 2015-01-30
Reviewed the reporting api guides

mnapoli:
#22

Not much here. I'm not that happy with that guide: http://developer.piwik.org/guides/querying-the-reporting-api It's really short with no explanations, only an example. But I don't have a lot of ideas right now. Maybe it would be worth to list client libraries (just like for the client libraries for the tracker API) in a separate guide.

### DIFF
--- a/docs/querying-the-reporting-api.md
+++ b/docs/querying-the-reporting-api.md
@@ -1,12 +1,13 @@
 ---
 category: Integrate
+previous: reporting-api-tutorial
 ---
 # Querying the Reporting API
 
 This guide explains how to call the Piwik API to request your web analytics data. There are two methods:
 
-*   using the standard HTTP API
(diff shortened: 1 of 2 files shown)


mattab:
> Maybe it would be worth to list client libraries (just like for the client libraries for the tracker API) in a separate guide.

def +1 to list the other third party Analytics API clients, there are some in Python, Node.js, Symfony2, CodeIgniter, Laravel. It makes sense in this page rather than assume our users use PHP. maybe more would follow if we start listing them more prominently this way :-)

mnapoli:
@mattab done in f593920e1efe4c82bd19cea1b7a877077b7f8024 and e74532b28a9847978dbfbfd1ed666448b064491a

PR #67, opened 2015-02-01
Documentation for the DI container

mnapoli:
I don't know if it's the right time to start documenting it. Feedback on the content is appreciated too.

### DIFF
--- a/docs/container.md
+++ b/docs/container.md
@@ -5,52 +5,123 @@ next: plugin-settings
 ---
 # Dependency Injection Container
 
-Some components in Piwik are configured and constructed using a **dependency injection container**.
+Most objects in Piwik are configured and constructed using a **dependency injection container**.
 
-## Retrieving from the container
+## Using Dependency Injection
 
-There are two ways to retrieve an entry (object or value) from the container:
+The general workflow for creating a new class when developing for Piwik is as follows:
 
-- **dependency injection**
-- getting it from the container using **`Container::get()`**
+1. **Determine what type of class you are creating and whether it should be stored in the container.**
 
-### Dependency injection
+  1. If you are creating a class that contains important application logic and depends on some part of Piwik, such as the Translator class or a DAO class, then it should be stored in the container.
 
-The preferred way to retrieve objects or values from the container is **to not call the container at all**.
+  2. If you are creating a class whose only purpose is to store data (such as a data structure or value object), it should not be stored in container.
 
-For example the `Translator` needs a `Loader` that loads translation files. Here is how it is retrieved:
+  3. If you are creating a class that can be used outside of Piwik, then you should use a third party library (or create a third party library) instead.
 
-```php
-class Translator
-{
-    private $loader;
+  4. If you are creating a class that is a combination of #1 and #2, then you should split it into two classes.
 
-    public function __construct(Loader $loader)
-    {
-        $this->loader = $loader;
-    }
+  In general, the rule is, if your class contains important application logic, it should be stored in DI.
 
-    // ...
-}
-```
+2. **If the class should be stored in DI, determine and specify its dependencies.**
 
-By using [dependency injection](http://fabien.potencier.org/article/11/what-is-dependency-injection), our class is easily testable and decoupled from the container. To instruct the container to inject the correct loader, we need to configure it (see the *Configuration* section).
+  A class' dependencies are the other classes stored in DI that it depends on. For example, an API class that serves data from the DB may depend on a DAO class. Create a list of these dependencies.
 
-### Container::get()
+  Then specify them as constructor arguments in your new class. For example:
 
-Another way to get entries from the container is to directly call `Container::get($id)`. This method is not ideal because it means we are coupling our code to the container and our class is also no longer unit-testable.
+  ```php
+  use Piwik\Translation\Translator;
+  use Piwik\Plugins\MyPlugin\Dao\MyEntityDao;
 
-However, given the current Piwik codebase is not following the pattern of dependency injection everywhere, it is accepted to call this method when dependency injection is not practical.
+  class API
+  {
+      /**
+       * @var Translator
+       */
+      private $translator;
 
-Example:
+      /**
+       * @var MyEntityDao
+       */
+      private $myEntityDao;
 
-```php
-$foo = StaticContainer::get('entry-name');
-```
+      public function __construct(Translator $translator, MyEntityDao $myEntityDao)
+      {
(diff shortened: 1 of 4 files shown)


mattab:
our developer guide would only showcase those classes tagged with `@api`, maybe we are ready to tag Container and StaticContainer with `@api` ? 

in any case this guide would be very useful even for anyone contributing to Piwik core or plugins, and of course to community when writing custom plugins.

mnapoli:
Your comment made me think about this and I actually wrote this doc for Piwik developers, not plugin developers.

I think we should hold off as long as we can exposing this to plugins, because as soon as it's exposed we have to keep BC. If it's only used in Piwik (and not 3rd party plugins) it leaves us some times to define best practices on naming container entries (in config files), on how to define entries, etc... We are just starting to use the container with plugins, we should gain more experience before freezing it with BC.

Also, `ContainerInterface` is a standard interface, not included in Piwik (and not related to PHP-DI). It doesn't make sense to tag it with `@api` since it's already an standard API (and it's not Piwik code anyway). `StaticContainer` is very bad practice, I don't want people using it, but at the same time Piwik's architecture doesn't leave us much of a choice… I'm conflicted.

So in conclusion, if we want the doc only for core developers we shouldn't merge it yet right?

mattab:
> So in conclusion, if we want the doc only for core developers we shouldn't merge it yet right?

Correct, the developer.piwik.org is for Piwik users and plugin developers, so let's not merge it for now

tsteur:
Now that we have a "Core Development" and a "Plugins / Develop" section can we write one guide for plugin developers under `/develop` Utils section that explains config and maybe DI to plugin developers. The guide should only contain information that is needed for plugin developers. And then an extended guide for core developers under `/core`? We could link from the short plugin guide to the extended guide if they want to know more.

diosmosis:
I updated the PR. @mattab @tsteur please review and let me know if you have questions/comments.

docs/container.md
```diff
@@ -0,0 +1,313 @@
...
+  * Implementations of the [Auth interface](http://developer.piwik.org/api-reference/Piwik/Auth).
+  * `Widgets` classes.
+  * Plugin `Settings` classes.
+  * `Type` classes.
+  * Plugin `Menu` classes.
+  * Plugin `Tasks` classes.
+
+  In 3.0, all classes will be constructed with dependency injection.
+
+### Configuring the container
```
tsteur: I think all the part above this section should go into a different page under http://developer.piwik.org/piwik-in-depth as it is quite complicated and not really super relevant for plugin developers.

docs/container.md
```diff
@@ -0,0 +1,313 @@
...
+We use dependency injection for the following reasons:
+
+* to move the responsibility of resolving dependencies outside of individual classes. Classes only list what they need, and the DI container gives it to them.
+* to allow plugins and users the ability to extend Piwik in any way they need by replacing and extending objects in the DI container.
+* to allow developers to more easily test classes both in isolation and in the Piwik environment, by making it easy to inject mocks.
+* to allow developers to more easily maintain code modularity by avoiding hard-coded dependencies.
+
+### What classes go in the container
+
+In Piwik development, there are four types of classes that exist:
```
tsteur: Is this part kinda repeating the first part? Maybe also rather something for  a page under http://developer.piwik.org/piwik-in-depth ?

docs/container.md
```diff
@@ -0,0 +1,313 @@
...
+  * `Widgets` classes.
+  * Plugin `Settings` classes.
+  * `Type` classes.
+  * Plugin `Menu` classes.
+  * Plugin `Tasks` classes.
+
+  In 3.0, all classes will be constructed with dependency injection.
+
+### Configuring the container
+
```
tsteur: I think for plugin developers relevant is 

* How to configure the DI container and how to replace dependencies etc (covered by this section)
* How to get an instance from DI container in the plugin itself (eg via constructor). Maybe we could also explain why one something has to use `StaticContainer` etc. Covered partially by the other sections but too detailed and too complicated I'd say. Maybe we can use that example `API` above to explain it.

Plugins should be pretty small usually and DI might be not too important for them. Eg they possibly don't need to create classes that can be changed by other plugins etc. This would be rather a use case for the core. I think for plugins more important is how can I change dependencies defined in core and how can I get the correct dependency eg for `Config`, `DB` in core

docs/container.md
```diff
@@ -0,0 +1,313 @@
+---
+category: Develop
+previous: piwiks-ini-configuration
+next: plugin-settings
```
tsteur: I'm not sure where this would currently appear as `piwiks-ini-configuration` is under "Piwik In Depth" and `plugin-settings` under "Develop > Utils". 

Maybe we could have
* one guide `Dependency Injection` under "Develop > Utils" explaining how to get dependencies
* one guide `Configuration` under "Develop > Plugin Basics" describing the config section from the guide below and all the possibilities one has. This could be as well part of the `Dependency Injection` guide. I'm not sure if it's beneficial to have 2 separate guides for this or whether it is confusing. We can still link between them so one can find it in 2 places which might be better. Not sure!
* and the more detailed documentation part in "Piwik In Depth > Utils" next or in the configuration page?

docs/container.md
diosmosis: This guide is mostly for core + PRO developers. 3rd party plugin developers do not actually have to use DI, and if they never write tests they're obviously never going to use DI. I can move the whole guide to piwik-in-depth or to some section for core/pro devs and revert to napoli's original guide for plugin devs.

> Eg they possibly don't need to create classes that can be changed by other plugins etc.

I know this is about plugin development, but I want to make it clear in case it's not: for core development (and probably pro development, though I don't know how much affect we have there), all classes that contain important application logic should be stored in the container. Regardless of whether there is a reason for them to be changed by plugins or not.

It is not each class' responsibility to obtain their dependencies (ie, by creating instances manually, accessing singletons or using a service locator), it is the container's responsibility to wire them. Any class that doesn't have important application logic could be replaced or stored in a 3rd party library or component.

docs/container.md
diosmosis: This sounds good to me. Will look into it when I have the time.

> I'm not sure if it's beneficial to have 2 separate guides for this or whether it is confusing. We can still link between them so one can find it in 2 places which might be better. Not sure!

I think it's ok, some redundancy in the writing (specific to each use case) could provide a greater chance the reader will understand.

tsteur:
documented in https://github.com/matomo-org/developer-documentation/pull/356 and targeted for plugin developers

PR #72, opened 2015-02-18
Improved develop page for plugin developers

tsteur:
I removed many links from the develop menu that are only important
for core developers or pro developers to a section named Core development
which can be reached from the develop menu (very bottom of the menu) or
from the develop-introduction page. We could maybe also name it
"Piwik in depth" or similar instead of "Core development" in case someone
is interested in those details.

I splitted some pages already in 2, eg tests-travis for plugin developers
and tests-travis for core.

Getting started is now only one page. The second part is now under Plugin
basics as "Custom Report". The third getting started page is completely
removed as it is explained in the other guides anyway.

The menu contained many links before that are not important for the majority
as we are only a few core developers and some of us do not even really use
developer.piwik.org as we are aware of this stuff. It should be still reachable.
That's why there are links to the core section and some guides might link to
the extended guides in core directly.

### DIFF
--- a/app/helpers/Content/Category/CategoryList.php
+++ b/app/helpers/Content/Category/CategoryList.php
@@ -13,6 +13,8 @@ class CategoryList
         switch ($name) {
             case 'Integrate':
                 return new IntegrateCategory();
+            case 'CoreDevelop':
+                return new CoreDevelopCategory();
             case 'Develop':
                 return new DevelopCategory();
             case 'Design':
--- a/app/helpers/Content/Category/CoreDevelopCategory.php
+++ b/app/helpers/Content/Category/CoreDevelopCategory.php
@@ -1,0 +1,61 @@
+<?php
+/**
+ * Piwik - Open source web analytics
+ *
+ * @link http://piwik.org
+ * @license http://www.gnu.org/licenses/gpl-3.0.html GPL v3 or later
+ */
+
+namespace helpers\Content\Category;
+
+use helpers\Content\EmptySubCategory;
+use helpers\Content\Guide;
+use helpers\Content\InternalLink;
+use helpers\Content\RemoteLink;
+
+class CoreDevelopCategory extends Category
+{
+    public function getName()
+    {
+        return 'Core Development';
+    }
+
+    public function getItems()
+    {
+        return [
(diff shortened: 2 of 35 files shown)


tsteur:
The menu for plugin developers does now fit on one screen! It would be nice to also see "Internationalization" under plugin basics but therefore we'd have to merge Internationalization / Localization + Multilingual which is some work. Might be also to much for one menu item we need to see in a next step! 

menu_for_plugin_developers


This is the new core developers menu... As it is unimportant to the majority it does not have a main menu item next to Intergrate and Develop

menu_for_core_developers

mattab:
nice changes! The goal here is to make the Developer docs focused on Plugin developers and KISS for them. Each guide should be  short and focused on one aspect of the platform (and most of them "use case oriented").

mnapoli:
Stats show that the "understanding piwik" kind of guides have a lot of views. Now they are hidden in a sub category that is listed like a guide, at the bottom of the long left menu, i.e. nobody will see them anymore. And it's especially more confusing since when clicking on that guide the left menu changes, but not enough to make us understand that we are actually in a new category. My first reaction was to think that actually everything in that subcategory was removed, and that the only thing left in "Core Development" (now Piwik in depth) was the 2 links in the empty page:

capture d ecran 2015-02-24 a 14 47 51

tsteur:
I can understand that many developers click on it as it appears quite top in the menu and "How Piwik Works" is quite normal that it is clicked. The information there is not really needed in order for plugin development but maybe nice to have for people who want to get more insights. It also contains content about Tracking API and Reporting API and not only develop related. 

Maybe we could move it to "Develop" after "Distributing your plugins". Otherwise we could maybe rename "Develop" to "Plugins", add a new category "Core" (where this is listed) and maybe remove "Home" in favor of the "Core" category.

I don't think "How Piwik Handles HTTP Requests" are important for plugin developers. Ideally they never have to use a controller and don't have to care about routing etc. This is only interesting for Core developers only and for people that want to contribute to Piwik. So maybe it is worth adding a category "Core" to the main menu. Then we'd have to tweak some guides maybe and clean up the menu. Eg we would move "Contributing to core" directly under "Introduction" (which we maybe would rename to "Getting started").

PR #75, opened 2015-02-19
Renamed Core Development Category to Piwik In Depth

tsteur:
* Renamed "PHP Documentation" section in API reference to "PHP Plugins API"
* Slightly moved "Utils" section in "Piwik In Depth" category further up for consistency with "Develop"


mnapoli:
> - Renamed "PHP Documentation" section in API reference to "PHP Plugins API"

:-1: I don't even understand the title anymore. This section is the phpdoc, it's the documentation of the PHP code. And it only includes Piwik Core it seems, there is no documentation of plugins in there.

> Renamed Core Development Category to Piwik In Depth

:-1: don't like that, it more confusing now.

tsteur:
`PHP Plugins API` maybe we could name it `PHP API for plugins`? It does not show the PHP documentation of all classes from PHP core, only the ones that we marked as API for plugins.

We can't name it `Core Development` as it is not only related to Core but mainly interesting for Core developers. We had several different names in mind and `Piwik in depth` seemed to be best. We can also name it differently.

mattab:
`PHP API for plugins` sounds more clear

`Piwik in depth` is quite messy section and applies both to Plugins and Core devs. Hard to find good name, but someone has an idea?

PR #81, opened 2015-02-22
Shorten the introduction of some guides

tsteur:
Some guides but only a few had an `About this guide` section. I think in the beginning most guides had such a section but now only a few have it and I suggest to remove it from the remaining ones under `Develop` as well. 

Most of them always have the same content eg you need to know PHP, read the getting started guide, ...

Guides should be written in a form that they only require the Getting started guide as a precondition. If one anyone needs special knowledge during the guide we can link to it. Eg to Less or Twig.

Some mentioned requirements were even kinda wrong or not ideal. Eg Translations guide mentioned one needs to have knowledge of PHP, Twig, JavaScript, ... and then there are 3 sections about PHP, Twig, JavaScript. One actually doesn't need knowledge of Twig if one is only interested in the PHP section for example. You get the point.

### DIFF
--- a/docs/security-in-piwik.md
+++ b/docs/security-in-piwik.md
@@ -3,21 +3,6 @@ category: Develop
 ---
 # Security best practices
 
-## About this guide
-
-**Read this guide if**
-
-* you'd like to know **how to write secure code when extending Piwik or contributing to Piwik**
-
-**Guide assumptions**
-
-This guide assumes that you:
-
-* can code in PHP, JavaScript and SQL,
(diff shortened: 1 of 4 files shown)


docs/security-in-piwik.md
```diff
@@ -3,21 +3,6 @@ category: Develop
 ---
 # Security best practices
 
-## About this guide
```
tsteur: The "Introduction" section pretty much contained the same as the "About this guide" section but much shorter and more clear. Eg it does not say "if you want you can do this" it says "you must" which is better.

Issue #87, opened 2015-03-10
Improve documentation for UI tests


mnapoli:
Document fixtures, etc.

PR #88, opened 2015-03-10
Refs #87, add UI tests doc section for Fixture use.

diosmosis:
@mnapoli Think this helps? Let me know if you have any thoughts or just merge if it's ok.


mnapoli:
Thanks! I thought the fixtures where actually in the zip file :p

PR #89, opened 2015-03-31
Explain how dataTable blobs are stored in the database

tsteur:
Thought it might be worth to explain it somewhere and to have it written down as it might be confusing.


mattab:
re-created in https://github.com/piwik/developer-documentation/pull/197

PR #94, opened 2015-05-13
Added information how to get plugins translated

sgiehl:
refs piwik/piwik#7883


mattab:
The workflow & new documentation looks good to me

PR #95, opened 2015-06-22
Added link where user can find Common constans

sebastianpiskorski:
Simple update as in title.


mattab:
cheers for this

Fyi, a github trick: in case the file in master will change the direct link may point to the wrong lines. For the next time we can write the link replacing `/master/`  by the tag name eg. https://github.com/piwik/piwik/blob/2.14.0-b8/core/Common.php#L26-L29 and this link will always work and point to right line

sebastianpiskorski:
nice trick, i will remember +1

mattab:
actually I listed the constant values in the guide since they won't change, and readers need the information. in https://github.com/piwik/developer-documentation/commit/7336ae707e0d981ab9321e990caa075251df7c0b

(also added the latest action types)

PR #99, opened 2015-07-28
Warning about the forced timezone change

LeCoyote:
The documentation should make it clear that including Piwik's index.php will forcefully change the default timezone to UTC. This is not mentioned anywhere at the moment.


mattab:
Thanks @LeCoyote  for improving the developer guide :+1:

PR #100, opened 2015-08-06
Documentation for dimensions

tsteur:
refs piwik/piwik#7837 : Deprecate no longer needed / wanted events 

This doc will be needed once we deprecate tracker events. It's not perfect but it is better compared to the event docs that we had before. Eg it is still missing what the next step is after creating a dimension. Eg how can one create a report and archive this data afterwards but truth is it is not possible yet.

I also noticed we need to improve some things in the API. Eg we should make it simpler to get params etc but will either make those changes directly or create an issue

### DIFF
--- a/docs/dimensions.md
+++ b/docs/dimensions.md
@@ -53,7 +53,7 @@ When a new visitor is being tracked, the value `1` will be stored. On each furth
 
 ## Action dimensions
 
-Action dimensions work similar, just the method name is differently. Let's say you want to track the current speed of a runner by exposing a new URL parameter `speed`:
+Action dimensions work similarly, just the method name is different. Let's say you want to track the current speed of a runner by exposing a new URL parameter `speed`:
 
 ```php
 public function onNewAction(Request $request, Visitor $visitor, Action $action)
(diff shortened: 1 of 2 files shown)


docs/dimensions.md
```diff
@@ -0,0 +1,83 @@
...
+    return $visitor->getVisitorColumn($this->column_name) + 1;
+}
+```
+
+When a new visitor is being tracked, the value `1` will be stored. On each further tracking request the value for this dimension will be increased.
+
+
+## Action dimensions
+
+Action dimensions work similar, just the method name is differently. Let's say you want to track the current speed of a runner by exposing a new URL parameter `speed`:
```
diosmosis: similar => similarly / just => only / differently => different

docs/dimensions.md
```diff
@@ -0,0 +1,83 @@
...
+A tracking request could be done like this: `piwik.php?idsite=1&sport_activity_type=running&speed=50`.
+
+Of course you can add any custom behaviour like limiting the max speed etc.
+
+
+## Segmentation
+A new segment is automatically created for a dimension and defined in the method `configureSegments`. All you usually have to do is to adjust the description of the accepted values.
+
+## Learn more
+Dimensions are quite powerful. For example you can change the behavior of an existing dimension by creating a dimension that has the same column name, you can store action related data very space efficient by using a lookup table and they can force the creation of a new visit in case an existing visitor was recognized. We recommend to have a look at the documentation within a created dimension and at the API-Reference of the classes [VisitDimension](/api-reference/Piwik/Plugin/Dimension/VisitDimension), [ActionDimension](/api-reference/Piwik/Plugin/Dimension/ActionDimension) and [ConversionDimension](/api-reference/Piwik/Plugin/Dimension/ConversionDimension)
```
diosmosis: 'very space efficient' => efficiently
'and they can force' => '. Dimensions can also force'

diosmosis:
Looks good to me. I can make the changes to the docs myself before or after merging if you like.

tsteur:
Thanks @diosmosis ! I made the changes and will merge

PR #106, opened 2015-10-19
Fix tracking code typos

merwan:
Added a trailing slash to the Piwik URL (like in the previous tracking code example) and removed the extra square bracket for site id.


mattab:
Great finds @merwan :+1:

PR #112, opened 2015-12-03
Documented tracking of Custom Dimensions

tsteur:
refs piwik/piwik#9129 

Kinda depends on https://github.com/piwik/plugin-CustomDimensions/issues/3 since we need a link for the Guide re how to install and configure Custom Dimensions


tsteur:
Once this is merged, we should update Custom Dimensions README and link to this instead of describing similar thing there (we'd have to keep too many places up to date). Ideally we would also link from the UI to this doc.

mattab:
:+1:

mattab:
> we should update Custom Dimensions README and link to this instead of describing similar thing there (we'd have to keep too many places up to date). Ideally we would also link from the UI to this doc.

@tsteur i'll let you do this

PR #120, opened 2016-02-16
Method doesn't exists

tassoman:
getInstance() method is now getContainer()


mattab:
Thanks @tassoman for the documentation fix!

PR #124, opened 2016-04-13
Small fixes to how-piwik-works

ihatehandles:


tsteur:
Thank you so much for this as well. Great to see improvements to the docs 👍

ihatehandles:
My pleasure. Actually worked on a couple more but didn't want to DDoS you with all of them at once before making certain my contributions weren't going the wrong way. So I'll send the other ones soon (almost midnight here).

tsteur:
Looking forward to them :)

PR #130, opened 2016-04-22
Touches to custom reports extended

ihatehandles:


tsteur:
Awesome, thanks again 👍 This is so appreciated

PR #136, opened 2016-08-02
Documented support fields

tsteur:
ping @mattab


mattab:
Looks good! if you wanted we could change `irc` to `chat` maybe, and `rss` to `feed` or so.

tsteur:
We could add these on top but I would not like to rename them as they are defined in composer like this. So developers "know" it out of the box if they define it anyway

mattab:
OK, makes sense to reuse the composer ones for consistency. LGTM

PR #140, opened 2016-08-15
when selecting "3.X" load latest version of the page rather than homepage

mattab:


mattab:
Note: not tested

mattab:
@tsteur the behavior is now consistent for all versions in the selector and will load the  page for the selected piwik version (rather than loading homepage)

app/routes/page.php
```diff
@@ -51,6 +51,7 @@ function renderGuide(Slim $app, Guide $guide, Category $category)
         'guide'              => $guide,
         'linkToEditDocument' => $guide->linkToEdit(),
         'activeMenu'         => $category->getName(),
+        'currentPath'        => $app->environment['PATH_INFO']
```
tsteur: I think this should be $app->request->getPathInfo()

Also global view variables need to be set in initView otherwise it may fail eg on home page, 404 page etc.

app/helpers/Twig.php
```diff
@@ -21,4 +22,21 @@ private static function getCompleteUrlFilter()
...
     }
 
+    private static function getCompleteUrlForPiwikVersionFilter()
+    {
+        return new \Twig_SimpleFilter('completeUrlForPiwikVersion', function ($value, $piwikVersion) {
+
+            $currentPiwikVersion = Environment::getPiwikVersion();
+
+            // we now work in context of that piwik version
+            Environment::setPiwikVersion($piwikVersion);
```
tsteur: Switching from 2.x to 3.x on home does not work

mattab:
Thanks for review, maybe now it's working? also fixed another small issue

PR #145, opened 2016-08-18
Document new preview field for plugin.json to define a demo url and video url

tsteur:


mattab:
Made minor changes and LGTM

PR #149, opened 2016-09-27
Ddded docs for pageviewid

tsteur:
refs https://github.com/piwik/piwik/pull/10499


mattab:
Now documented @ http://developer.piwik.org/api-reference/tracking-api

PR #150, opened 2016-10-02
Make text bold within code blocks for better contrast and readability

mattab:
I reckon it looks better.


mattab:
Note: also made this change on piwik.org / marketplace as well

PR #151, opened 2016-10-03
Added a license

redalertexpert:
Hey guys,

I saw you can just find the license of your project in the readme file. I would say, that this is not thaaat good style :wink: , at least since GitHub features licenses next to contributors. 

Cheers


mattab:
The code & content is not under GPL actually... it's copyright. it's not allowed to copy or replicate it. I'll remove license from readme, thanks for the note.

PR #152, opened 2016-10-12
Mark the 'require' field as required as it is now required

mattab:


mattab:
Thanks  @danielstonies for letting us know

PR #160, opened 2017-01-12
Add docs for jwplayer and videojs

tsteur:


docs/media-analytics/custom-player.md
```diff
@@ -8,7 +8,7 @@ In this guide we will learn how you can implement [Media Analytics](http://www.m
  
 ## Tracking a new media player in the browser
 
-The Media Analytics plugin currently supports  out of the box the HTML5, Vimeo and YouTube media players. If you are using
+The Media Analytics plugin currently supports out of the box the HTML5, Vimeo and YouTube media players. If you are using
```
mattab: Could we also here mention JW Player, VideoJS ? as people don't necessarily know that these players use html5

tsteur: I had it first here but then removed it again. That's why there is the space change. I'd personally rather not have it here. They are all based on HTML5 and we actually support many more players etc. We can add it, I just prefer not to

mattab:
:+1: 

Will need to add it to User guide and FAQs too

tsteur:
already done a 5 hours ago

PR #175, opened 2017-02-24
Tracking multiple domain names in the same website

mattab:
Refs https://github.com/piwik/piwik/issues/2211

```
To record users across multiple domain names in the same Piwik website, we need to setup what is called Cross Domain linking. Cross domain tracking in Piwik makes sure that when the visitor visits multiple websites and domain names, the visitor data will be stored in the same visit and that the visitor ID is reused across domain names. A typical use case where cross domain is needed is for example when an ecommerce online store is on www.awesome-shop.com and the ecommerce shopping cart technology is on another domain such as secure.cart.com.

Learn about setting up cross-domain tracking in our guide: How do I accurately measure a same visitor across multiple domain names (cross domain linking)?
```

### DIFF
--- a/docs/3.x/tracking-javascript-guide.md
+++ b/docs/3.x/tracking-javascript-guide.md
@@ -499,6 +499,11 @@ _paq.push(['setDomains', '*.example.com']);
 
 _paq.push(['trackPageView']);
 ```
+### Tracking multiple domain names in the same website
+
+To record users across multiple domain names in the same Piwik website, we need to setup what is called Cross Domain linking. Cross domain tracking in Piwik makes sure that when the visitor visits multiple websites and domain names, the visitor data will be stored in the same visit and that the visitor ID is reused across domain names. A typical use case where cross domain is needed is for example when an ecommerce online store is on `www.awesome-shop.com` and the ecommerce shopping cart technology is on another domain such as `secure.cart.com`.
+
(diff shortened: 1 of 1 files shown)


docs/3.x/tracking-javascript-guide.md
```diff
@@ -499,6 +499,11 @@ _paq.push(['setDomains', '*.example.com']);
 
 _paq.push(['trackPageView']);
 ```
+### Tracking multiple domain names in the same website
+
+To record a visitor across different domain names into a single visit within one Piwik website, we need to setup what is called Cross Domain linking. Cross domain tracking in Piwik makes sure that when the visitor visits multiple websites and domain names, the visitor data will be stored in the same visit and that the visitor ID is reused across domain names. A typical use case where cross domain is needed is, for example, when an ecommerce online store is on `www.awesome-shop.com` and the ecommerce shopping cart technology is on another domain such as `secure.cart.com`.
+
+Learn about setting up cross-domain tracking in our guide: [How do I accurately measure a same visitor across multiple domain names (cross domain linking)?](https://piwik.org/faq/how-to/faq_23654/)
```
tsteur: Not sure re title "Tracking multiple domain names in the same website".

wouldn't this also require "setDomains()" etc? Maybe the title should be like "How to identify the same visitor across domains" or so?

tsteur: Otherwise 👍

PR #183, opened 2017-05-27
fix typo

t3chguy:


sgiehl:
Thanks for the PR. There is actually the same typo in Piwik. I've proposed a PR for that.
Btw. the same typo still exists in line 414

t3chguy:
@sgiehl are you sure?
> Both `checkOnScroll` and `timeIntervalInMs` cannot be changed after this method was called the first time.

sgiehl:
sorry, checked that on my local copy. Guess I forgot to pull ;)

mattab:
Thanks for the PR @t3chguy - appreciated :+1:

PR #184, opened 2017-06-21
Document getPiwikUrl and getCurrentUrl

mattab:
from https://github.com/piwik/piwik/pull/11616/files


mattab:
Note: we should also document these ones:


* A new JavaScript tracker method getNumTrackedPageViews has been added to retrieve the number of tracked page views within the currently loaded page or web application.
* New JavaScript tracker methods setSessionCookie, getCookie, hasCookies, getCookieDomain, getCookiePath, and getSessionCookieTimeout have been added for better cookie support in plugins.

PR #185, opened 2017-06-22
Note concerning testing of heartbeattimer, see piwik/piwik#11808

larsnystrom:
See piwik/piwik#11808.

I just think it'll help people to explicitly mention dev tools.


mattab:
Thanks for the note!

PR #189, opened 2017-07-31
Update help about where the auth_token can be found.

letharion:
I had trouble finding the auth_token based on the existing help, so I made an attempt at clarifying.


mattab:
Thanks for the PR @letharion :+1:

PR #195, opened 2017-10-12
tracking API 'url' parameter is actually not required

mattab:
reported in https://github.com/piwik/piwik-sdk-ios/issues/194


brototyp:
@mattab You moved the line but still left the `(required)` in there. I think you want to change this to a `(recommended)` instead.

mattab:
Nice catch, will fix this now :

PR #198, opened 2017-10-24
Fix search

### LINKED ISSUE #23: Searching for user guides does not always work
Reproduce:
- Search for `contrib` 
- Got: no result instead.

Findus23:
Fixes #23 

It got a bit more than just one change as the old typeahead version didn't support the "all" parameter and the new one works a bit different.

But altogether the search now works as it should and the json file is loaded only once on focus of the search box.


app/public/js/app.js
```diff
@@ -24,42 +24,36 @@ $('.documentation table').addClass('table table-striped table-bordered');
...
+    $quickSearchTypeahead.on("focus", function () {
+        var url = $quickSearchTypeahead.attr('data-action');
+        $.get(url, {}, function (quickSearchData) {
+            $quickSearchTypeahead.typeahead({
+                source: quickSearchData.names,
+                items: 'all',
+                updater: function (item) {
+                    // get text to display in quick search box
+                    var displayText = item;
+                    console.log(item);
```
mattab: probably want to remove the `console.log`

Findus23: You are right, I removed it in https://github.com/piwik/developer-documentation/pull/200/commits/a163bae1e3379a5d0feda1b3dbee366aea6bb542 and forgot to also do so here.

But keep in mind that this function doesn't exist any more with the full text search in #200 as I am using a maintained fork of twitter bootstrap there, because it's more user friendly (results can be selected with the arrow keys)

Findus23:
Closing in favor of #330

PR #200, opened 2017-10-26
Modernization (Slim 3, Bootstrap 4, working search)

### LINKED ISSUE #23: Searching for user guides does not always work

### LINKED ISSUE #27: links with anchor such as /reporting-api#Overlay do not show the title
Steps to reproduce:
- Open  http://developer.piwik.org/api-reference/reporting-api#Overlay 
- Got: title `Overlay` is behind top bar
- expected: visible `Overlay` below top bar

title hidden

### LINKED ISSUE #97: Developer Website is slow
Just opened the website and noticed the website is pretty slow. It takes over 4s to load and the JS and font files are never cached. The site used to load in under 500ms. It's also a bit annoying since sometimes some content is there but it takes 2 or 3 more seconds until the titles are rendered (because they use a custom font now). My internet connection is 2Mbit/s, testing with Chrome.

JS and fonts should be loaded from cache, ideally they would be loaded async. 

developer website

Findus23:
replaces #198 
fixes #23 and fixes #97 and fixes #27 
I had more time than expected, so I went quicker than expected.

# Changes
- use npm to fetch javascript libraries and bower for compiling CSS and concatenate all libraries in a single file
- ~~change the logic of the search~~
  - ~~send data with JSON mime-type~~
  - ~~fetch it only once before initializing the search~~
- rewrite all routes and middleware to work with Slim 3
- as Slim 3 doesn't include logging use monolog for logging
- update the markdown parsing library (I didn't notice any changes)
- replace Bootstrap 3 with Bootstrap 4
- use sass (scss) for CSS
- update CSS so the style stays about the same
  - made search in navbar a bit more obvious
- keep same caching, ~~but also cache the /data/documents route~~
- add basic opengraph data (for forum)
- full text search
  - exclude 2.x and generated doc
  - in addition still search through titles and phpdoc as full text search doesn't prioritize them
  - generate index on webhook
  - keep (modified) old index for mappings from markdown file to URL and title
- add authentification to webhook
  - While a GET parameter isn't perfect it's a simple solution for keeping random people from running the (quite resource hungry) tasks
  - `/receive-commit-hook?token=changeme` 
# TODO
- I am not sure if the logging works as intended. I can't think of many errors that would result in slim exceptions instead of php errors
- I haven't yet tested the styling in different browsers. And I am not sure if it will work in IE < 11. But I don't know if there are many developers using outdated browsers.
- Maybe using a memory based cache can improve the performance

# Instance with default settings:
https://issues.lw1.at/


tsteur:
@Findus23 how far are you here? Is it possible to merge this?

tsteur:
Maybe we merge some things now, and add more things in other PRs? To not have one big PR in the end, to minimize risk of breaking things, and to deploy things sooner...

Findus23:
I'd say this is mostly finished.

In the next days winter holidays start and I'll have more time to look into it again.
Splitting it up may be difficult as quite a few parts depend on each other, but I'll try to create this parts:
- php changes (slim).
- css changes
- full text search
- single commits that fix unrelated issues I came across

tsteur:
same here @Findus23 are you still working on this? Can we do anything?

Findus23:
This would need a bit more work.
After merging the PHP part in #213, I'd recreate the move to bootstrap 4 
The search part I'd still keep out as I am least happy about it.

mattab:
@Findus23 is there a chance you could schedule this work over the next few weeks maybe?

PR #202, opened 2017-10-30
Adds instructions for viewing UI demo page

spacenate:
Also links to these instructions from other relevant guides.

Fixes piwik/piwik#11710, possibly also piwik/piwik#7652.


tsteur:
Awesome, thanks for this 👍

PR #203, opened 2017-10-31
Improve database schema doc for Users and Permissions

mattab:


mattab:
Merging now but feel free to review / suggest improvements anyway I will apply them after

PR #204, opened 2017-11-02
Remove no longer needed articles and add angular guide to in depth section

tsteur:
see https://github.com/piwik/developer-documentation/issues/199#issuecomment-341228736


docs/3.x/angularjs-getting-started.md
```diff
@@ -39,11 +44,11 @@ I recommend watching some videos and read a few resources about the benefits and
   * Some directives such as ng-if and ng-repeat create their own scope
   * All bound data should be dotted when using ng-model otherwise you are doing it wrong. You would be wondering why a value is not updated and it will take you hours to figure this out.
     Example:
-    ```<div ng-model="model.searchfield">``` instead of ```<div ng-model="searchfield">```
+    `&lt;div ng-model="model.searchfield">` instead of `&lt;div ng-model="searchfield">`
```
mattab: Maybe this won't be displayed properly in markdown ie. will be displaye as html entities?

mattab:
Left a comment otherwise LGTM :+1:

docs/3.x/angularjs-getting-started.md
tsteur: Nope, it was before shown wrong, then I fixed to the entities to make sure it is shown properly.

tsteur: Funny, I had to change it after all. Locally it works like this because I remember that I needed to change it to make it work. I must be using an outdated composer snapshot and an older version of the markdown lib or so.

PR #212, opened 2017-12-12
fix sidebar active class

Findus23:
When one visits https://developer.piwik.org/api-reference/reporting-api-metadata, two items are highlighted as https://developer.piwik.org/api-reference/reporting-api is a substring


Findus23:
I also thought that it wouldn't work. Maybe it was for a page that doesn't exist any more.

PR #213, opened 2017-12-12
Upgrade to Slim 3

Findus23:
Cherry-picked all changes from #200 that don't affect JS/CSS/Twig Templates. Also doesn't include the full text search.

~~I'll test if this breaks anything soon.~~
## To do after/while merge:
- run composer install as this updates some libraries
- (just to be save) delete the twig cache in app/tmp/templates/


app/helpers/CacheMiddleware.php
```diff
@@ -16,30 +19,33 @@
...
 
             if (!empty($content)) {
-                $res->setBody($content);
-
-                return;
+                if ($this->isJsonData($req)) {
+                    $res = $res->withHeader('Content-Type', 'application/json');
+                    $res->getBody()->write($content);
+                } else{
+                    $res->getBody()->write($content . "\n<!-- Cached response -->");
```
tsteur: is this comment fine to append to a JSON content type?

Findus23: I don't think so. That's why I'm only appending it when it isn't a json document.

tsteur: Stupid me :) didn't look at it properly

app/helpers/CacheMiddleware.php
```diff
@@ -64,15 +69,22 @@ private function pathToCacheKey($path)
...
-    private function shouldCache($req)
-    {
+    private function shouldCache(Request $req) {
         return $req->isGet();
     }
 
-    private function getCacheKey($req)
-    {
+    private function isJsonData(Request $req) {
+        return strpos($req->getUri()->getPath(), 'data/') !== false;
```
tsteur: I haven't tested it, but what if the URL contains `data/` as in `https://developer.piwik.org/guides/log-data/xxx`? This URL wouldn't work but there might be something in the future. I haven't looked yet at the structure of `data` URLs but maybe we can check if it starts with a certain URL or something else? Just a thought...

Findus23: You are right, that could cause issues in the future. I’ll replace it with something more specific

Findus23: This should be fixed now in 667cc0b04960e9f1387eb24d9a5216738928270d

Findus23:
@tsteur 
Can you also check if the logging works the way you want?
Slim 3 doesn't include logging anymore and I am not sure if I have understand how logging in PHP/monolog work exactly.

tsteur:
@Findus23 so far everything seems to work but the logger as you mentioned. Is there a chance we can somehow keep the `Log` class so it won't fail here for example: https://github.com/piwik/developer-documentation/blob/master/app/helpers/Markdown/IncludeFilePostprocessor.php#L73

Findus23:
@tsteur 
I thought I have already answered...

The log class doesn't work anymore as Slim 3 doesn't include any logging feature anymore, but recommends to use monolog instead. Unfortunately I don't know how to access the class in the `IncludeFilePostprocessor.php`. 
For now I have commented the line out and added a check so the website doesn't get cached if an `file_get_contents()` error occurs.

tsteur:
@Findus23 what's the current status here?

Findus23:
I think this is the branch where I extracted the PHP changes and it should be working. (apart from logging)
I can go through everything again next month, but after testing, this should be fine to merge.

mattab:
@Findus23 Be awesome if we could merge your work soon and benefit from the improvements :+1:

tsteur:
🚀

PR #217, opened 2018-01-11
Clarify available license terms and new policy for removing outdated plugins

tsteur:
refs https://github.com/matomo-org/matomo/issues/12429


mattab:
As discussed internally we've agreed that:

>  6 months after a major Matomo version is no longer supported, we will remove any plugin version from the Marketplace that supports only this no longer supported Matomo version. For example when Matomo 2.X reaches its end of life in December 2017 (receives no longer any updates nor security fixes), we will remove all plugin versions that target only Matomo 2.X in June 2018. This may cause a plugin to disappear from the Marketplace when there is no longer any up to date version. Please note it is always recommended to update to a supported Matomo version in order to still receive updates and security fixes.

PR #219, opened 2018-01-31
Add new rule for marketplace plugins to reuse existing libraries

tsteur:


mattab:
👍 good to clarify this aspect in the guidelines

PR #237, opened 2018-05-09
Replace 'last' with 'previous'

jmumby:
'last' has been used in the 'previous' example


mattab:
Thanks for the PR @jmumby and congratulations on having your first PR merged :tada:

PR #243, opened 2018-05-14
Add docs for creating a custom opt out form.

diosmosis:


Findus23 (approved):
I haven’t tested it but the docs are easy to understand.

mattab:
@diosmosis :+1: I tested it in a simple use case and it worked.

(Noticed that our Iframe opt-out does not read the 1st party cookie, but that's quite good as it makes sure the iframe always checks if the opt-out is "global" rather than just for one domain. maybe we should mention this in this documentation?)

diosmosis:
Don't think the iframe opt-out is able to read the 1st party cookie since it will be on the matomo domain. Makes sense to mention this somewhere 👍 ( can make the change (eventually) if I know where)

PR #245, opened 2018-05-26
Don't use jQuery for custom opt-out

Findus23:
As most developer documentation doesn't depend on jQuery and this example doesn't use anything complex, it should also be working without jQuery.

Only thing that's a bit modern is `querySelector`, which is also supported by 97.6% of all users:
https://caniuse.com/#feat=queryselector

See also https://forum.matomo.org/t/dsgvo-deutschsprachige-opt-out/28424/8?u=lukas

### DIFF
--- a/docs/3.x/tracking-javascript-guide.md
+++ b/docs/3.x/tracking-javascript-guide.md
@@ -890,29 +890,28 @@ Below is a jQuery-based example opt-out form that replicates the built in Matomo
   </p>
 </div>
 <script>
-jQuery(function ($) {
-  function setOptOutText() {
-    _paq.push([function () {
-      $('#optout').attr('checked', this.isUserOptedOut() ? undefined : 'checked');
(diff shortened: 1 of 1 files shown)


docs/3.x/tracking-javascript-guide.md
```diff
@@ -890,29 +890,28 @@ Below is a jQuery-based example opt-out form that replicates the built in Matomo
...
-    setOptOutText();
+    setOptOutText(optOut);
   });
-
-  setOptOutText();
+  setOptOutText(optOut);
 });
-</script>
-```
+</script>```
```
tsteur: don't know if the ```
need to be on a new line?

tsteur:
Haven't tested the script but looks good to merge to me. Added one comment.

Findus23:
@tsteur Thanks for the note. I fixed the ``` and removed the `jQuery-based`

PR #246, opened 2018-05-27
Remove the example on how to call the API through PHP

tsteur:
This is not needed and HTTP should always be recommended. There is not really a performance or resources improvement over going through HTTP. Also the example doesn't work and would need to be updated regularly. None of those methods mentioned there are API and it is only a matter of time until it breaks (it did in the past eg when we added environment etc).

### DIFF
--- a/docs/3.x/querying-the-reporting-api.md
+++ b/docs/3.x/querying-the-reporting-api.md
@@ -5,10 +5,7 @@ next: reporting-api-clients
 ---
 # Querying the Reporting API
 
-This guide explains how to call the Piwik API to request your web analytics data. There are two methods:
-
-- using the standard HTTP API 
-- using Piwik's PHP code directly
+This guide explains how to call the Piwik API to request your web analytics data. 
 
 ## Call the Piwik API using the HTTP API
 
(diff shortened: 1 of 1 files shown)


docs/3.x/querying-the-reporting-api.md
```diff
@@ -32,22 +29,6 @@ Here is the output of this code:
...
 ``` 
 
-## Call the Piwik API in PHP
-
-If you want to request data in a PHP script **that is on the same server as Piwik**, you can use this simple technique. This is a more efficient solution as it doesn't require network calls. You directly call the PHP Piwik runtime and get the PHP data structure back.
-
-If you are developing a plugin, you should be using this technique. Please note that including index.php will set the default timezone of the calling code to UTC.
-
-```php
-{@include escape https://raw.github.com/matomo-org/matomo/master/misc/others/api_internal_call.php}
```
sgiehl: Guess we could remove the included file then as well?

tsteur: Sounds good to me, I can issue a PR later. It seems to be used in a test but not quite sure why.

PR #250, opened 2018-06-25
Document new methods to disable the recording of mouse movements

tsteur:


tsteur:
Once merged, we need to create an FAQ like this: https://matomo.org/faq/heatmap-session-recording/faq_24215/

PR #256, opened 2018-07-27
improve docs about setting up tests

Findus23:
Feedback from https://secure.helpscout.net/conversation/626522262/119343/

I feel like the tests docs should be further expanded by someone who successfully runs the tests locally.


Findus23:
Not required thanks to #259 and @johsin18.

PR #277, opened 2018-11-13
Remove docs for Piwik 2

### LINKED ISSUE #276: A few 404 errors for 2.x documentation pages that don't actually exist
There are maybe 20 or more similar 404 pages found by googlebot

On this page for example: https://developer.matomo.org/api-reference/Piwik/Settings/FieldConfig#ui_control_password

Click on "Piwik 2.x" leads to 404:  https://developer.matomo.org/2.x/api-reference/Piwik/Settings/FieldConfig

Edit: solution could be to remove the links to the piwik 2.x docs when the classes / code doesn't exist there?

tsteur:
According to our policy we remove plugins and docs 6 months after we ended support for that version. For Piwik 2.X this was in June 2018 so it is time to remove the docs.

This PR removes the selector in the top menu.

@mattab when opening eg https://developer.matomo.org/2.x/guides/distributing-your-plugin should we redirect to https://developer.matomo.org/guides/distributing-your-plugin or should we still show the old content which shows the link to the new version?

fix https://github.com/matomo-org/developer-documentation/issues/276


Findus23:
https://www.google.com/search?q=site%3Adeveloper.matomo.org shows hardly any 2.x sites, so I'd say people who go there do it intentionally to get the old docs.

tsteur:
> I reckon it is fine to remove the old docs 12 months after ending support for a particular version, and then redirect to the latest docs.
> Also it would be great to make the warning more clear on the page, maybe a yellow/warning system, and also leverage this warning message to also invite people to upgrade their Matomo soon?

Can you create issues for those things? For now I would just merge it and keep things simple. And I would make up too many rules and not make things complicated with 6 months after end we remove it but still show content, then after 12 months we start redirecting etc. The docs for 2.x and 3.x are mainly the same and for sure will be for 3.x and 4.x Not sure there's much of a point to keep it for so many years. This PR will keep the content when accessed directly so should be all good to merge for now.

Issue #289, opened 2019-01-29
update 404 github links


dev-101:
couple of 404s:
https://developer.matomo.org/guides/theming

![dev matomo_github_theme_404_links_](https://user-images.githubusercontent.com/5383059/51903047-ab38e580-23bb-11e9-96a5-092a8faa9a6b.png)

PR #291, opened 2019-01-29
Update theming guide

tsteur:
fix https://github.com/matomo-org/developer-documentation/issues/289

### DIFF
--- a/docs/3.x/theming.md
+++ b/docs/3.x/theming.md
@@ -21,34 +21,40 @@ After you enter the appropriate information, a theme will be created for you in 
 
 ## Simple theming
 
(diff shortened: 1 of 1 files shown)


Findus23:
@tsteur Maybe we should further explain which variables should be set in PHP and which in LESS (or link to a source listing them.

When creating a theme recently I only made changes to the LESS variables.
https://github.com/Findus23/matomo-ClassicTheme/blob/master/stylesheets/theme.less

Findus23:
@tsteur So I can set every variable (even specific ones like `graph-colors-data-series1`) via PHP?

tsteur:
Graph colors could be explained as part of the advanced guide 👍  It's already partially doing it. Or they could be added to `themeStyles` in core.

PR #299, opened 2019-03-26
replace idSite=7 with idSite=62

Findus23:
idSite=7 (I think it was the forum) isn't public anymore so all API examples should now use idSite=62
@diosmosis Is it okay to run the API examples against the demo site? And is 62 keeping the same id?

There are more broken examples that should be fixed:

- https://demo.matomo.org/?module=API&method=ImageGraph.get&idSite=62&apiModule=CustomVariables&apiAction=getCustomVariables&token_auth=anonymous&period=day&date=2013-11-11,2013-11-18&flat=1&filter_pattern_recursive=.*logged.* is showing the fraction of logged in users on the forum which now doesn't work anymore

- the bulk tracking example doesn't work: 
https://demo.matomo.org/?module=API&method=API.getBulkRequest&format=xml&urls[]=module%3DAPI%26method%3DVisitorInterest.getNumberOfVisitsPerVisitDuration%26format%3DXML%26idSite%3D62%26period%3Dday%26date%3D2013-11-24%26expanded%3D1&urls[]=module%3DAPI%26method%3DUserSettings.getBrowser%26format%3DXML%26idSite%3D7%26period%3Dday%26date%3D2013-11-24%26expanded%3D1


diosmosis:
It's fine to run the examples against the demo site, but they may need some tweaked parameters since the data will be different. The idSite won't change.

PR #304, opened 2019-05-20
Update UI test instructions to use Puppeteer

katebutler:


docs/3.x/tests-ui.md
```diff
@@ -214,7 +222,7 @@ The page renderer object has the following methods:
 - **mousedown(selector, [waitTime])**: Sends a mouse down event to the element referenced by `selector`.
 - **mouseup(selector, [waitTime])**: Sends a mouse up event to the element referenced by `selector`.
 - **sendKeys(selector, keyString, [waitTime])**: Clicks an element to bring it into focus and then simulates typing a string of keys.
-- **sendMouseEvent(type, pos. [waitTime])**: Sends a mouse event by name to a specific position. `type` is the name of an event that phantomjs will recognize. `pos` is a point, eg, `{x: 0, y: 0}`.
+- **sendMouseEvent(type, pos. [waitTime])**: Sends a mouse event by name to a specific position. `type` is the name of an event. `pos` is a point, eg, `{x: 0, y: 0}`.
```
sgiehl: actually most of those methods have changed, as they are used directly from puppeteer. I think none of them supports `[waitTime]` any longer, methods like `load` are now named `goto`. Mouse events are called `page.mouse.move()`,...
Not sure if the list here makes much sense, or if we should refer to the Puppeteer docs. ping @diosmosis

diosmosis: Yes, probably simplest to just refer to puppeteer docs saying we forward to that API. Only thing we added is `waitForNetworkIdle()` which is also automatically called on page loading functions. Also `screenshotSelector()`, which can be an alternative if normal puppeteer screenshots fail randomly.

tsteur: @katebutler could you look at the comments here?

katebutler: Have removed the outdated list of methods and updated all JS code snippets to use Puppeteer. Also added docs for `persist-fixture-data` as there were none currently.

mattab:
Merging cc @diosmosis @tsteur @sgiehl if you spot anything please open a new PR :+1:

PR #310, opened 2019-09-30
Matomo Tag Manager to track Single Page Application

mattab:
* Adding a little section about info for SPA and Tag Manager
* Question: do we maybe need to also recommend users to create a Custom HTML tag with the following code (or similar), to reset the state correctly, when new pages are loaded within a SPA?

```
    // remove all previously assigned custom variables, requires Piwik 3.0.2
    _paq.push(['deleteCustomVariables', 'page']); 
    _paq.push(['setGenerationTimeMs', 0]);
    _paq.push(['trackPageView']);
    
    // make Matomo aware of newly added content
    var content = document.getElementById('content');
    _paq.push(['MediaAnalytics::scanForMedia', content]);
    _paq.push(['FormAnalytics::scanForForms', content]);
    _paq.push(['trackContentImpressionsWithinNode', content]);
    _paq.push(['enableLinkTracking']);
```

### DIFF
--- a/docs/3.x/spa-tracking.md
+++ b/docs/3.x/spa-tracking.md
@@ -136,7 +136,8 @@ If you're using [Tag Manager](https://matomo.org/tag-manager/) to implement your
 
 To trigger your Matomo tag (which calls `trackPageView`), you can either:
 1. use the "History change" [trigger](https://matomo.org/docs/tag-manager/#triggers) which would work in most cases,
-2. or in your Single Page App, you can trigger a Tag Manager Event `{event: 'mtm.PageView'}` by calling the following line in JavaScript: `_mtm.push({'event': 'mtm.PageView'});`
+2. or in your Single Page App, if you are using the 'Pageview Trigger' to trigger a Pageview, you can trigger a Tag Manager Event `{event: 'mtm.PageView'}` by calling the following line in JavaScript: `_mtm.push({'event': 'mtm.PageView'});`. 
+   - This would also work similarly when you use instead the 'DOM Ready Trigger' (call `_mtm.push({'event': 'mtm.DOMReady'});`) or when you use the 'Window Loaded Trigger' (call `_mtm.push({'event': 'mtm.WindowLoad'});`
 
 
 
--- a/docs/3.x/spa-tracking.md
+++ b/docs/3.x/spa-tracking.md
@@ -129,6 +129,17 @@ window.addEventListener('hashchange', function() {
     _paq.push(['enableLinkTracking']);
 });
 ```
+
+## Using Matomo Tag Manager to track Single Page Application
+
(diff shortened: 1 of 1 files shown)


docs/3.x/spa-tracking.md
```diff
@@ -129,6 +129,17 @@ window.addEventListener('hashchange', function() {
...
 });
 ```
+
+## Using Matomo Tag Manager to track Single Page Application
+
+If you're using [Tag Manager](https://matomo.org/tag-manager/) to implement your Matomo Analytics Tracking, then in your Single Page Application,  whenever there is a new page loaded in your app you will need your [Matomo Tag](https://matomo.org/docs/tag-manager/#configuring-a-tag) to be triggered for the Page view to be tracked. 
+
+To trigger your Matomo tag (which calls `trackPageView`), you can either:
+1. use the "History change" [trigger](https://matomo.org/docs/tag-manager/#triggers) which would work in most cases,
+2. or in your Single Page App, you can trigger a Tag Manager Event `{event: 'mtm.PageView'}` by calling the following line in JavaScript: `_mtm.push({'event': 'mtm.PageView'});`
```
tsteur: Should we add something like `If you are using the PageView trigger do trigger a Pageview`. They could also be using DomReady or PageLoaded event etc?

tsteur:
Yes would say they would need to do that @mattab not sure if this can be embedded eventually into tag manager. Especially the "scan" calls etc

mattab:
@tsteur updated the doc, please review/merge if correct?

docs/3.x/spa-tracking.md
```diff
@@ -129,6 +129,18 @@ window.addEventListener('hashchange', function() {
...
 ```
+
+## Using Matomo Tag Manager to track Single Page Application
+
+If you're using [Tag Manager](https://matomo.org/tag-manager/) to implement your Matomo Analytics Tracking, then in your Single Page Application,  whenever there is a new page loaded in your app you will need your [Matomo Tag](https://matomo.org/docs/tag-manager/#configuring-a-tag) to be triggered for the Page view to be tracked. 
+
+To trigger your Matomo tag (which calls `trackPageView`), you can either:
+1. use the "History change" [trigger](https://matomo.org/docs/tag-manager/#triggers) which would work in most cases,
+2. or in your Single Page App, if you are using the 'Pageview Trigger' to trigger a Pageview, you can trigger a Tag Manager Event `{event: 'mtm.PageView'}` by calling the following line in JavaScript: `_mtm.push({'event': 'mtm.PageView'});`. 
+   - This would also work similarly when you use instead the 'DOM Ready Trigger' (call `_mtm.push({'event': 'mtm.DOMReady'});`) or when you use the 'Window Loaded Trigger' (call `_mtm.push({'event': 'mtm.WindowLoad'});`
```
tsteur: @mattab I think `DOMReady ` and `WindowLoad ` don't have the mtm prefix

PR #320, opened 2019-12-20
Update docs since resetUserId no longer creates new visit

### LINKED ISSUE #15293: New visit not created after logout
FYI, In our application we are using UserId feature.

I am setting below attributes when user clicks on logout button in our Application.
_paq.push(['resetUserId']);
_paq.push(['trackPageView']);

But, when the user logs back new visit is not created for the user with same UserId.

Please let me know, what is wrong here?

tsteur:
fix https://github.com/matomo-org/matomo/issues/15293

cc @mattab can you have a look?

### DIFF
--- a/docs/3.x/tracking-javascript-guide.md
+++ b/docs/3.x/tracking-javascript-guide.md
@@ -406,6 +406,10 @@ _paq.push(['resetUserId']);
 _paq.push(['appendToTrackingUrl', 'new_visit=1']); 
 
 _paq.push(['trackPageView']);
+
+// we make sure to not again create a new visit afterwards (important for Single Page Applications)
+_paq.push(['appendToTrackingUrl', '']); 
+
 ```
 
 
--- a/docs/3.x/tracking-javascript-guide.md
+++ b/docs/3.x/tracking-javascript-guide.md
@@ -391,17 +391,19 @@ _paq.push(['trackPageView']);
 _paq.push(['enableLinkTracking']);
 ```
 
-### When user logs out, reset User ID
+### When user logs out, reset User ID 
 
-*(resetUserId is available since Matomo 3.3.1)*
+When the user has logged out and a User ID is not available anymore, it is recommended to notify Matomo by calling the `resetUserId` method before `trackPageView`. 
 
-When the user has logged out and a User ID is not available anymore, it is recommended to notify Matomo by calling the `resetUserId` method before `trackPageView`:
+If you want to create a new visit when your users logout, then you can also force Matomo to create a new Visit:
 
 
 ```javascript
 
-// User has just logged out, we reset the User ID which also generates a new Visitor ID
+// User has just logged out, we reset the User ID
 _paq.push(['resetUserId']);
+// we also force a new visit to be created for the pageviews after logout
+_paq.push(['appendToTrackingUrl', 'new_visit=1']); 
 
 _paq.push(['trackPageView']);
 ```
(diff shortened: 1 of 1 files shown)


mattab:
The doc still says: `// User has just logged out, we reset the User ID which also generates a new Visitor ID`

Although the code is:
```

            /**
             * Clears the User ID and generates a new visitor id.
             */
            this.resetUserId = function() {
                configUserId = '';
            };

```

Can we change the comment to `// User has just logged out, we reset the User ID which un-sets the User ID` or so?

mattab:
Or should we actually maybe rather reset the Visitor ID, when `resetUserId` is called, to indeed force a new visit to be created? This way we don't have to update https://matomo.org/faq/how-to/faq_187/ and break BC?

tsteur:
@mattab I don't think that's a good idea. It's like then you can as well undo the entire change. Also AFAIK we are not doing this when you set a userId so it's not quite consistent behaviour. It would be kind of taking control out of the user if we always created a new visit then maybe... could be done though... but then you'd also need to create a new visit when setting a userId etc.

mattab:
@tsteur slightly tweaked to document how to force creating a new visit. 
I'm now wondering, what is the point of calling `resetUserId` at all when users don't call the `['appendToTrackingUrl', 'new_visit=1']` method?

mattab:
also added another commit after reading https://github.com/matomo-org/matomo/issues/14739#issuecomment-519268164

diosmosis:
> what is the point of calling resetUserId at all when users don't call the ['appendToTrackingUrl', 'new_visit=1'] method?

👍 i don't think that makes sense either. have to do both to associate the action with an entirely new context (w/o associating it w/ any userid).

tsteur:
> what is the point of calling resetUserId at all when users don't call the ['appendToTrackingUrl', 'new_visit=1'] method?

The point is consistent behaviour. Currently, we are not creating a new visit when someone logs in. So it be inconsistent to create a new visit when the user logs out. See https://github.com/matomo-org/developer-documentation/pull/320#issuecomment-568628719

Behaviour should be either consistent where it creates a new visit on log in, and log out. Or not at all.

diosmosis:
> The point is consistent behaviour. Currently, we are not creating a new visit when someone logs in. 

But the docs here say we should create a new visit when logging someone out? Maybe I'm misreading the conversation on the ticket?

tsteur:
@diosmosis Yes. I was initially creating this doc to mention it no longer creates a visit as this is what is implemented. It could be changed though to create a new visit on log in and log out.

mattab:
FYI: I've also changed the guide:

from:
> Signing-out use case: In the case where a visitor visits your website and is logged-in (User ID is set) then a visit will be created for this User ID. If then visits logs out then the visitor has no User ID set any more and a new visit will be created on the page view right after logging out. (if User ID was not used then the requests would have been tracked into one visit and one unique visitor). Starting in Matomo 3.4.0 we introduce a new functionality to record a user logging out, via the method resetUserId, learn more about about When user logs out, reset User ID in the developer guide.

to:

> Signing-out use case: In the case where a visitor visits your website and is logged-in (User ID is set) then a visit will be created for this User ID. If the user then logs out and has no User ID set any more (`resetUserId` function is called), a new visit will not be created for the page views after logging out. Subsequent pageviews will still be tracked in the existing visit, and this visit will still have the User ID set. You can also easily force a new visit to be created when users log out, learn more in the developer guide for JavaScript Tracking: [when user logs out, reset User ID](https://developer.matomo.org/guides/tracking-javascript-guide#when-user-logs-out-reset-user-id)

mattab:
FYI also made another change to call deleteCookies in https://github.com/matomo-org/developer-documentation/pull/325/files which is now consistent with https://matomo.org/faq/how-to/faq_187/

PR #325, opened 2020-01-23
When user logs out, call deleteCookies to also force a new Visitor ID

mattab:


tsteur:
Why would you want a new visitor ID? I thought the whole point would be to keep the visitorID.... 
Be good to not merge PRs yourself. Thanks.

PR #327, opened 2020-01-29
adding "for Apps or Server-Side tracking" to the "Client libraries" header

mattab:
also creating a FAQ where i'll link to this sub-header directly


mattab:
merging now as I need the link anchor but let me know if any suggestion

PR #328, opened 2020-02-04
Upgrade to Slim 4

Findus23:
Third time's a charm :)

This is a recreation of #213 using the latest version of Slim 4 that again changed quite a lot of things.

This only includes the most important changes to get Slim 4 working, any other improvement to the site will come in later pull requests to keep this small.

- updated Slim to 4
- updated Twig to 3.x
- use latest michelf/php-markdown
- requires at least PHP 7.2.0
- also cache json routes
- use proper file extensions for cache files
- don't cache if page contains failed request
- cache routing to `tmp/cache/route_cache.php`
- completely rewritten ` app/public/index.php ` as it contains the slim-specific logic
- lots of reformatting in `app/routes/page.php` to have routes in new format
- partly revert #277 (show search again in navbar)
- minor change in `app/public/js/app.js` to make search work again.

I'm still totally confused by Dependeny Injection so this is written in a "normal" style instead of as recommended in https://github.com/slimphp/Slim-Skeleton/blob/master/public/index.php.

A demo page running this branch can be found at https://matomodocs.lw1.at/


tsteur:
👍

PR #329, opened 2020-02-05
require token for webhook

Findus23:
recreation of 36e65247be01aa69bb267c6f8e5102defbedd24e

Allowing everyone to clear the cache allows to put a huge load on the server. 

Therefore a custom token should be defined in the local.php and the webhook config should be changed to use `/receive-commit-hook?token=thetoken`


tsteur:
Configured it. Haven't tested it yet on the server but we'll see if things don't update automatically anymore

tsteur:
I'm getting expected response though when testing

PR #330, opened 2020-02-05
use bootstrap3-typeahead for search

### LINKED ISSUE #23: Searching for user guides does not always work

Findus23:
recreation of #198

makes the search far more user-friendly (e.g. keyboard controls work)

also fixes #23

again a demo can be seen on https://matomodocs.lw1.at/


tsteur:
Cheers 👍

PR #333, opened 2020-02-05
support api-reference subdirectories

Findus23:
fixes https://developer.matomo.org/api-reference/tagmanager/javascript-api-reference

I didn't notice this section was added since I last rewrote the routes.


tsteur:
works now 👍 Thanks

PR #334, opened 2020-02-23
Document changed behaviour of heartbeat timer

### LINKED ISSUE #15612: document how heart beat timer works since 3.13.0
https://github.com/matomo-org/matomo/pull/15058 made it work more efficiently, but now it isn't intuitive anymore as one can't see it do its thing when just quickly testing the website.

Therefore at least https://developer.matomo.org/guides/tracking-javascript-guide#accurately-measure-the-time-spent-on-each-page and maybe also related FAQs need to be updated to properly describe how it works and how to test if it is set up correctly.

tsteur:
fix https://github.com/matomo-org/matomo/issues/15612

### DIFF
--- a/docs/3.x/tracking-javascript-guide.md
+++ b/docs/3.x/tracking-javascript-guide.md
@@ -148,7 +148,7 @@ It is possible to configure Piwik so that it accurately measures the time spent 
 _paq.push(['enableHeartBeatTimer']);
 ```
 
-Piwik will then send requests to count the actual time spent on the page, when the user is actively viewing the page (ie. when the tab is active and in focus). The heart beat request is executed when:
+Piwik will then send requests to count the actual time spent on the page, when the user is actively viewing the page (i.e. when the tab is active and in focus). The heart beat request is executed when:
  * switching to another browser tab after the current tab was active for at least 15 seconds (can be configured see below)
  * navigating to another page within the same tab. 
 
--- a/docs/3.x/tracking-javascript-guide.md
+++ b/docs/3.x/tracking-javascript-guide.md
@@ -148,11 +148,13 @@ It is possible to configure Piwik so that it accurately measures the time spent 
 _paq.push(['enableHeartBeatTimer']);
 ```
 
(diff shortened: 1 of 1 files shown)


docs/3.x/tracking-javascript-guide.md
```diff
@@ -148,11 +148,13 @@ It is possible to configure Piwik so that it accurately measures the time spent
 _paq.push(['enableHeartBeatTimer']);
 ```
 
-Piwik will then send requests to count the actual time spent on the page, when the user is actively viewing the page (ie. when the tab is active and in focus). These heartbeat requests will not track additional actions or pageviews. By default, Piwik will send a heartbeat request every `15` seconds. You may change the default interval, for example to send a request every `30` seconds:
+Piwik will then send requests to count the actual time spent on the page, when the user is actively viewing the page (ie. when the tab is active and in focus). The heart beat request is executed when:
```
Findus23:
```suggestion
Piwik will then send requests to count the actual time spent on the page, when the user is actively viewing the page (i.e. when the tab is active and in focus). The heart beat request is executed when:
```
While you are on it (even though I think it should be e.g. instead)

Findus23 (approved):
Sounds good, thanks!

docs/3.x/tracking-javascript-guide.md
```diff
@@ -148,11 +148,13 @@ It is possible to configure Piwik so that it accurately measures the time spent
 _paq.push(['enableHeartBeatTimer']);
 ```
 
-Piwik will then send requests to count the actual time spent on the page, when the user is actively viewing the page (ie. when the tab is active and in focus). These heartbeat requests will not track additional actions or pageviews. By default, Piwik will send a heartbeat request every `15` seconds. You may change the default interval, for example to send a request every `30` seconds:
+Piwik will then send requests to count the actual time spent on the page, when the user is actively viewing the page (i.e. when the tab is active and in focus). The heart beat request is executed when:
+ * switching to another browser tab after the current tab was active for at least 15 seconds (can be configured see below)
+ * navigating to another page within the same tab.
```
dakur, line 153: @tsteur @Findus23 I'm just having tough time understanding what Matomo's heartbeat actually does. It is stated it sends requests when user is actively viewing the page, but in next sentence it is written that is executed only on window blur. So what's true? According to https://github.com/matomo-org/matomo/pull/15058 the latter, but the name `heartbeat` suggests it makes requests in some interval…

PR #336, opened 2020-03-25
Consent for Safari users: consent only remembered 7 days

mattab:
I also wanted to write `If the user visits again within 7 days, the cookie will be renewed. If the user visits more than 7 days after the previous visit, then they will need to give their consent again. (Safari only)` but it's not currently a feature. 

@tsteur maybe it would make sense to update  our JS tracker and automatically renew the consent cookie on each visit? to minimise number of times consent will be asked for safari users, improve user experience on websites using Matomo and asking for consent.


tsteur:
@mattab feel free to create an issue. Not sure it's best from privacy perspective. Also we'd need to check if that's actually fine by Safari but would say so. Also in general it is currently a feature that a user can configure how long the consent should be valid for because they might be required to (or they want to for good privacy) ask for consent again every X days or months. This feature would then be broken and would need to be adjusted as well through a different flag and the cookie format would need to be migrated etc. Right now we don't really want to do this by default as it's maybe not the best from a privacy perspective.

tsteur:
BTW it seems a workaround could be to get the website to re-invalidate all cookies. Eg should PHP be used, it could do something like this:

```php
foreach ($_COOKIE as $name => $value) {
            if (strpos($name, '_pk') === 0) {
                setcookie($name, $value, time() + (3600 * 24 * 365), $path = '/', $domain = 'enteryourdomain.com', $secure = true, $httpOnly = true);
            }
        }
```

Above code seems to set the expiry date for 1 year. I tested in Safari 13.1 and worked for me on HTTPS. Ideally it should add samesite value though eg from PHP 7.3

```php
setcookie($name, $value, [
            'expires' => time() + (3600 * 24 * 365 * 1),
            'path' => '/',
            'domain' => 'enteryourdomain.com',
            'secure' => true,
            'httponly' => true,
            'samesite' => 'None',
        ]);
```

sgiehl:
@mattab @tsteur should that safari note still be added? If so, guess we need to recreate the PR as that content was moved to a new file in #360

tsteur:
The content for this file moved to it's own consent page. Moved the note over to the correct file.

PR #346, opened 2020-05-14
disable route cache during development

Findus23:


app/public/index.php
```diff
@@ -51,7 +51,9 @@
 $app->add(new CacheMiddleware());
 
 $routeCollector = $app->getRouteCollector();
-$routeCollector->setCacheFile('../tmp/cache/route_cache.php');
+if (!DEBUG) {
```
tsteur, line 54: @Findus23 should it be rather `CACHING_ENABLED` ?

Findus23: I was not sure, which one would be better. But if it is easier for you this way, I will change it.

PR #349, opened 2020-05-28
Support Matomo 4 [WIP]

tsteur:
Left to is:

* Copy pictures from the previous version into latest version eg from public/img/3.x to public/img/4.x and update all needed pictures.
* Copy `docs/3.x-dev` to `docs/4.x-dev`

We'll do this closer to the beta to not having to maintain both 3.x and 4.x docs for the next few weeks.


tsteur:
FYI Had to update our Sami to support PHP 7.2 syntax in the docs generator... it actually fixes few issues but in the hooks page it doesn't format variables correctly when they are references. That's just a minor visual thing.

PR #353, opened 2020-06-18
Document few changes for Matomo 4

tsteur:
refs https://github.com/matomo-org/matomo/issues/12420

Updated some docs based on the changelog and also updated some domains, piwik.js and matomo.js mentions etc.

### DIFF
--- a/docs/4.x/ab-tests/browser.md
+++ b/docs/4.x/ab-tests/browser.md
@@ -107,7 +107,7 @@ _paq.push(['AbTesting::create', {
       }
       return false;
   },
-  matomoTracker: Matomo.getAsyncTracker(piwikUrl, piwikSiteId),
+  piwikTracker: Matomo.getAsyncTracker(matomoUrl, matomoSiteId),
   variations: [
         // [...]
         {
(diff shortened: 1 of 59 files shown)


tsteur:
@sgiehl  or @Findus23 if someone could have a very quick look that be great.

docs/4.x/ab-tests/browser.md
```diff
@@ -107,7 +107,7 @@ _paq.push(['AbTesting::create', {
       }
       return false;
   },
-  matomoTracker: Matomo.getAsyncTracker(piwikUrl, piwikSiteId),
+  piwikTracker: Matomo.getAsyncTracker(matomoUrl, matomoSiteId),
```
sgiehl, line 110: Guess the `piwikTracker` option hasn't been renamed in the plugin yet?

sgiehl (approved):
LGTM

Findus23 (approved):
Didn't see anything weird when scrolling through. Maybe once we have replaced all piwik occurrences, we could remove the automatic replacement as I think it still breaks some pages.

docs/4.x/ab-tests/browser.md
tsteur: exactly, will change the premium feature docs once we renamed things there

PR #354, opened 2020-06-18
Add Matomo 4 Plugin Migration guide

tsteur:

Had a look through the changelog https://github.com/matomo-org/matomo/blob/4.x-dev/CHANGELOG.md but couldn't find many things that changed in particular and are especially interesting for plugin developers. Most of the changes are rarely used events and mostly used internally.

@diosmosis  @sgiehl @Findus23 let me know if you think anything needs to be explained in more detail compared to the changelog. Do we maybe have a list of the upgraded libraries? They be good to mention, although maybe rather in the changelog itself.

We can also always merge for now and add more things over time.


diosmosis:
There are some materialize changes that people may need to make, though if they use matomo directives, not as many. Should this be mentioned?

tsteur:
@diosmosis not sure. Do you have an example? eg a link to a diff or so? Guess also only directives mentioned here would need document if something changes there. https://github.com/matomo-org/matomo/blob/4.x-dev/tests/UI/expected-screenshots/Theme_demo.png as others aren't really considered API. Eg the form directives, piwik-block etc.

docs/migrate-matomo-3-to-4.md
```diff
@@ -0,0 +1,79 @@
...
+# Migrate Plugin from Matomo 3.X to Matomo 4
+
+This migration guide covers explains how to do some migrations to make a plugin compatible with Matomo 4. A list of
+all changes in Matomo 4 can be found in the [Changelog](/changelog).
+
+## Adjust the required Matomo version
+
+For your plugin to be executed in Matomo 4 you first need to configure it as compatible with Matomo 4:
+
+* specify that your plugin requires Matomo 4 (thr require for Piwik 3 used to be eg `"piwik": ">=3.0.0-b1,<4.0.0-b1"`).
```
Findus23, line 13:
```suggestion
* specify that your plugin requires Matomo 4 (the require for Piwik 3 used to be eg `"piwik": ">=3.0.0-b1,<4.0.0-b1"`). 
```

docs/migrate-matomo-3-to-4.md
```diff
@@ -0,0 +1,79 @@
...
+* The sameas and divisibleby tests are deprecated in favor of same as and divisible by respectively.
+
+e.g. `divisibleby(2)` becomes `divisible by(2)`
+
+## Summary
+
+In this guide we have seen which steps to take to migrate your Matomo plugin to be compatible with our latest Matomo 4.
+If you need further help for converting your plugin to Matomo 3, head over to the [Piwik developers community forums](https://forum.matomo.org/c/plugins-platform).
+
+Once you've adjusted your plugin don't forget to release a new version.
```
Findus23, line 79:
```suggestion
Once you've adjusted your plugin, don't forget to release a new version.
```

Findus23 (commented):
Just two minor typos. The rest sounds good.

docs/migrate-matomo-3-to-4.md
```diff
@@ -0,0 +1,79 @@
...
+category: Develop
+---
+# Migrate Plugin from Matomo 3.X to Matomo 4
+
+This migration guide covers explains how to do some migrations to make a plugin compatible with Matomo 4. A list of
+all changes in Matomo 4 can be found in the [Changelog](/changelog).
+
+## Adjust the required Matomo version
+
+For your plugin to be executed in Matomo 4 you first need to configure it as compatible with Matomo 4:
```
diosmosis, line 11: would reword this from "configure it as compatibile" to "show it is compatible with matomo 4 in your plugin.json file"

tsteur: 👍 changed

tsteur:
I'll merge for now and we can look at this again when we do the RC or so.

PR #366, opened 2020-08-06
archiving docs changes

diosmosis:
* [X] first draft
* [x] revise draft

Refs https://github.com/matomo-org/matomo/issues/16261

### DIFF
--- a/docs/4.x/archiving.md
+++ b/docs/4.x/archiving.md
@@ -79,7 +79,7 @@ Archive data is calculated differently based on the period type:
 - "day" periods are aggregation of log data
 - "week", "month", "year" and custom date ranges are aggregation of "day" reports
 
-For example archive data for a week is created by aggregating archive data of the 7 days of the week. This is much faster than aggregating log data.
+For example archive data for a week is created by aggregating archive data of the 7 days of the week. This is much faster than aggregating seven days of log data.
 
 ### Plugin Archivers
 
@@ -124,7 +124,7 @@ Persisted reports and metrics are indexed by the website ID, period and segment.
 The status of each archive is stored as a row in the `archive_numeric_*` tables. The archive status row has `name` value of
 `done*`, where the suffix can contain a specific plugin name and/or segment hash.
 
-In the code, these rows are called the "done flag"s of an archive, and the value of this row (called the "done flag value")
+In the code, this row is called the **done flag** of an archive, and the value of this row (called the **done flag value**)
 is the status of the archive. This row can have the following values:
 
 * `ArchiveWriter::DONE_OK` - the archive was successfully processed and can be read.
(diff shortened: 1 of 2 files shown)


Findus23 (approved):
Sounds good to me (can't talk about the content)

diosmosis:
FYI @tsteur merged this

docs/4.x/archiving.md
```diff
@@ -41,21 +41,24 @@ While pre-computing archive data seems of course superfluous for 1000 page views
 
 ## When?
 
-By default, archive data is calculated and cached **on-demand**. When a specific report is requested, Piwik will check if the required archive data exist and generate it if not.
+### On-demand Archiving
+
+By default, archive data is calculated and cached **on-demand**. When a specific report is requested, Piwik will check if the required archive data exists and generate it if it does not.
```
tsteur, line 46: Maybe we also add something like this:

If the report was archived recently (within configured amount of seconds in time_before_today_archive_considered_outdated) then no new archive is generated.

docs/4.x/archiving.md
```diff
@@ -41,21 +41,24 @@ While pre-computing archive data seems of course superfluous for 1000 page views
...
 
 ### Pre-archiving
 
-When tracking a website with a lot of traffic, the archiving on-demand might take too much time. In those situations, archiving on demand must be disabled and [pre-archiving needs to run in background at a scheduled time](https://matomo.org/docs/setup-auto-archiving/).
+When tracking a website with a lot of traffic, on-demand archiving will take too much time and resources, causing users to wait a long time before a report becomes visible.
```
tsteur, line 50: Maybe we could be more specific and say something like:

if you have more than 500 page views a day, or a slow server, or generally find Matomo slow, then we recommend you disable browser archiving and set up CLI archiving...

docs/4.x/archiving.md
```diff
@@ -137,10 +153,165 @@ For example, the *UserSettings* plugin uses one record to hold browser details o
...
+    // ...
+
+    public function __construct(ArchiveProcessor $processor)
+    {
+        parent::__construct($processor);
+
+        // if a single report is requested, mark the archive we're creating as partial
+        $this->requestedReport = $processor->getParams()->getArchiveOnlyReport();
+        if ($this->requestedReport) {
+            $processor->getParams()->setIsPartialArchive(true);
```
tsteur, line 285: @diosmosis didn't notice this was needed? Any chance we could handle this automatic? (`$processor->getParams()->setIsPartialArchive(true);`)

docs/4.x/archiving.md
```diff
@@ -137,10 +153,165 @@ For example, the *UserSettings* plugin uses one record to hold browser details o
...
+    }
+
+    public function aggregateMultipleReports()
+    {
+        if ($this->isArchiving('MyPlugin_mySpecificReport')) {
+            $this->getProcessor()->aggregateDataTableRecords(['MyPlugin_mySpecificReport']);
+        }
+    }
+
+    private function isArchiving(string $reportName)
```
tsteur, line 312: should we put this method maybe in the parent class so not every one has to add it?

tsteur (commented):
@diosmosis left few minor comments 👍

docs/4.x/archiving.md
diosmosis: No, because the extra code in aggregateDay/MultiplePeriods has to exist too, otherwise we'd mark a full archive as partial. It has to be done by Archivers that specifically support this behavior.

docs/4.x/archiving.md
diosmosis: 👍

docs/4.x/archiving.md
diosmosis: 👍

docs/4.x/archiving.md
diosmosis: Yeah, though if it's not used in conjunction w/ the other code, it'll just introduce bugs. I thought it was better to require users to be explicit about how they implement partial archiving support, and not just use it because they see it.

docs/4.x/archiving.md
tsteur: OK. I thought it would in worst case only end up with only a done flag in the archive but no actual report (or all reports archived again instead of the specifically requested report in that plugin).

docs/4.x/archiving.md
tsteur: OK no big preference. Just would avoid needing it in several archivers (since we'll add it to funnels, usersflow, custom reports, ...) but it's also not complex method so can be added into the archiver again

Issue #370, opened 2020-08-16
Update recommended coding standards on the "Contributing to Mamoto Core" page


nina-py:
This page: https://developer.matomo.org/guides/contributing-to-piwik-core has the following lines:

> Make sure your code follows the PSR 1, PSR 2 and PSR-4 coding standards.

On following the link to the PSR 2 standard, I see that it's been deprecated for just over a year:

> Deprecated - As of 2019-08-10 PSR-2 has been marked as deprecated. PSR-12 is now recommended as an alternative.

Solution: link to the PSR-12 standard instead?

tsteur:
Sounds good @nina-py  👍  do you mind creating a PR?

nina-py:
> Sounds good @nina-py +1 do you mind creating a PR?

Not at all, will do asap.

nina-py:
Thanks Thomas for merging the pull request!

PR #372, opened 2020-08-17
Issue #370: updated coding guidelines link

nina-py:
Updated coding guidelines to link to PSR 12 instead of the deprecated PSR 2.


tsteur:
Thank you very much @nina-py

PR #373, opened 2020-08-17
Trying to document &convertToUnicode

mattab:
*   **convertToUnicode**; by default (when set to 1), the output is sent in `UTF-16LE`. If set to 0, then the output will be encoded in `UTF-8`.

refs https://github.com/matomo-org/matomo/issues/16166#issuecomment-665430577


tsteur:
@mattab you would need to mention this only applies to CSV format?

PR #374, opened 2020-08-20
Update travis command to include most recent parameters

### LINKED ISSUE #16260: Update docs for running tests on travis
We need to update the docs in https://developer.matomo.org/guides/tests-travis to mention bionic distribution in the command for 4.x-dev (which I think is not used by default) and maybe in https://developer.matomo.org/guides/migrate-matomo-3-to-4 we could also mention they might need to run the generate travis yml command to run tests against 7.2 and 7.4?

I basically mean eg `--php-versions="7.2,7.4" --distribution="bionic" --sudo-false --verbose`

tsteur:
fix https://github.com/matomo-org/matomo/issues/16260

### DIFF
--- a/docs/4.x/tests-travis.md
+++ b/docs/4.x/tests-travis.md
@@ -17,7 +17,7 @@ Current status for master branch: [![Build Status](https://travis-ci.org/matomo-
 Plugins can do the same if they include a `.travis.yml` file in their github repository. You can generate this file using the `generate:travis-yml` console command:
 
 ```
-$ ./console generate:travis-yml --plugin=MyPlugin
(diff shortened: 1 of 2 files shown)


tsteur:
@sgiehl I'll merge for now. Let me know if anything needs updating

sgiehl:
Guess should be fine. The `--verbose` is actually not needed, but doesn't hurt either.
Additionally when updating it might be needed to update the exclude matrix, as that isn't done automatically and thus old an then unused PHP versions are excluded

tsteur:
Cheers @sgiehl made some tweaks in https://github.com/matomo-org/developer-documentation/commit/7fa9c7d2d7e293d174e05b77781c078f459f75c0

PR #378, opened 2020-08-27
Use \DI\autowire instead of \DI\object

sgiehl:
refs https://github.com/matomo-org/matomo/pull/16311


tsteur:
@sgiehl does `DI\link` also need to be updated? Not sure if there an example with a callback in there already but if not be good to add one with `DI\value(function...)` and one maybe where function is used as a factory?

sgiehl:
@tsteur wasn't able to find `DI\link` in Matomo, nor in any of our plugins or the documentation. Shall we mention it nevertheless?
Also added a block about adding event listeners

PR #380, opened 2020-09-04
Document setVisitStandardLength JS tracker function

mattab:
Added in https://github.com/matomo-org/matomo/pull/15058/files

probably also need to mention the function in https://matomo.org/faq/general/faq_36/ ?


tsteur:
Updating the FAQ too sounds good @mattab

PR #387, opened 2020-10-22
Using the same container on multiple websites

mattab:
new sub section as we got this request from a user and found this page to be the best place to document it


mattab:
fyi @tsteur

PR #393, opened 2020-11-03
Update tracking-javascript-guide.md

Starker3:
Added: 
Testing the hearbeat timer can be done by right clicking anywhere on the web page and clicking "Inspect Element" and then switching to the "Network" tab so see any requests being sent to Matomo

### DIFF
--- a/docs/4.x/tracking-javascript-guide.md
+++ b/docs/4.x/tracking-javascript-guide.md
@@ -159,7 +159,7 @@ Piwik will then send requests to count the actual time spent in the visit, as lo
 // Requires a page to be actively viewed for 30 seconds for any heart beat request to be sent.
 _paq.push(['enableHeartBeatTimer', 30]);
 ```
-
+Testing the hearbeat timer can be done by right clicking anywhere on the web page and clicking "Inspect Element" and then switching to the "Network" tab so see any requests being sent to Matomo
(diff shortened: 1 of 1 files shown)


Findus23:
I think it would be more straightforward to say something like: "You can check if the heartbeat timer is working correctly by checking the network tab of your browser developer tools (<kbd>F12</kbd>/`<kbd>F12</kbd>`) to see if the requests to matomo.php are sent correctly."

tsteur:
@Starker3 it's not that easy because you wouldn't see those requests directly because requests aren't simply sent eg every X seconds. It's waiting for certain actions like you clicking on a link (and then it's important to have "preserve log" across page views enabled ) or need to focus other windows etc. Could we instead create an FAQ on matomo org and then just link to it? I'd prefer to not have it in there as it adds unnecessary information for most users and could just link to the FAQ if someone really wants to know this and then explain in detail (since it's not trivial)

Starker3:
@tsteur We can do that instead, - @mattab wanted me to add it here since we had a question about it and the only reference we have concerning the Heartbeat Timer code was in the developer docs.
I'll write up an FAQ about it and review it with Matthieu when he is available

tsteur:
Great. Thanks @Starker3

PR #394, opened 2020-11-05
remove section about setGenerationTimeMs

Findus23:
This function was removed in https://github.com/matomo-org/matomo/pull/15736

@sgiehl I'd consider that a breaking change and we should probably mention it in the CHANGELOG.md

On my website Matomo 4 throws an error in the console as the function is still called.

### DIFF
--- a/docs/4.x/spa-tracking.md
+++ b/docs/4.x/spa-tracking.md
@@ -37,20 +37,6 @@ _paq.push(['trackPageView']);
 
 Note: we recommend you use [Custom Dimensions](https://matomo.org/docs/custom-dimensions/) instead of Custom Variables as they will be deprecated in the future.
 
-### Updating the generation time
-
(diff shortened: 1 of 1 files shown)


sgiehl (approved):
Guess it would make sense to also mention that in the change log. Feel free to create a PR for that.

Findus23:
@sgiehl Do you recommend any replacement or is there any real need for it anymore? How would the new timing work in a SPA?

sgiehl:
@Findus23 actually there shouldn't be any need for a replacement as the performance metrics should be gathered automatically. But we could document the new method `disablePerformanceTracking` if that wasn't done yet.

Findus23:
I guess this can be closed after https://github.com/matomo-org/developer-documentation/commit/234fcfcfcd2d1c78883dff05a163b4d0e74ef1ab and https://github.com/matomo-org/matomo/issues/16757

tsteur:
Sorry @Findus23 didn't notice there was a PR open for this

tsteur:
fyi documented the method in https://github.com/matomo-org/developer-documentation/commit/60b053f4c3dc647e7f35a3aefaf5ba0cb0836546

PR #396, opened 2020-11-16
Use new demo server to generate API list

mattab:


docs/4.x/reporting-api.md
```diff
@@ -213,4 +213,4 @@ All the commonly used terms in Piwik Analytics are documented in the [Glossary](
 
 ## API Method List
 
-{@include https://demo2.matomo.org/?module=API&action=listAllMethods&prefixUrl=https://demo.matomo.org/&idSite=62}
+{@include https://demo2.matomo.org/?module=API&action=listAllMethods&prefixUrl=https://demo.matomo.cloud/&idSite=1}
```
tsteur, line 216: @mattab might be better to use demo.matomo.cloud also as URL? What happened here is that now you are fetching plugins and API doc generator from Matomo 4 that points to a Matomo 3 so some links don't work (like PagePerformance) and it adds some parameters that may not be supported yet etc.

mattab: Makes sense. fixed in https://github.com/matomo-org/developer-documentation/commit/97307a50f4ba79b1a34527bda45952742876eeb2

 I checked the diff and indeed there are some plugins also that probably clutter the doc a bit like LogViewer, LoginSaml, CustomTranslations, Bandwidth, Provider, TreemapVisualization (which are now removed)

Btw @tsteur on the listing it now shows "Cloud" plugin which maybe we could hide from there?

PR #399, opened 2020-11-30
Fix link to milestones in core-team-workflow.md

AntiSol:
Correct link to milestones on github (should be https://github.com/matomo-org/matomo/milestones rather than https://github.com/matomo-org/matomo/issues/milestones)


tsteur:
Thank you @AntiSol very appreciated

PR #402, opened 2020-12-02
start on archiving behavior spec doc

diosmosis:
WIP will be adding more to it over time


tsteur:
@diosmosis maybe we could merge this one already or are you still working on it?

diosmosis:
@tsteur there are a lot of empty sections. So far it's been the lowest priority for me, but I can re-prioritize if needed.

diosmosis:
@matomo-org/core-reviewers can you all read this and see what makes sense and doesn't make sense? It's not an interesting read since it's the inner details and expectations of the archiving process, but it's an important process so it should be documented.

docs/4.x/archiving-behavior-specification.md
```diff
@@ -0,0 +1,385 @@
...
+
+`core:archive` will automatically invalidate the archives for `today` and `yesterday` if certain criteria are met.
+
+For `today`, the archive will be automatically invalidated if there are visits for today and if the existing archive is
+older than the configured TTL (see [the previous section on archiving settings](#settings)).
+
+For `yesterday`, the archive will be automatically invalidated if the current date is a different day than the `ts_archived`
+date for the latest archive for `yesterday`. If this is true, it means the day has changed since the archive was processed,
+and there may be more visits to process. For example, if the last known archive for 2021-05-20 is calculated at 2021-05-20 20:00:00,
+but it is now 2021-05-21 00:30:00, there may be visits between 20:00:00 and 00:00:00 (midnight) to process.
```
sgiehl, line 284: Is that only true for `yesterday`? If archiving doesn't run for some days, I guess this will also be done for an older day, right?

diosmosis: Currently only true for 'yesterday'. We assume core:archive runs at least once per day.

docs/4.x/archiving-behavior-specification.md
```diff
@@ -0,0 +1,385 @@
...
+the entrypoint for report aggregation.
+
+### General Expected Behavior
+
+API methods use the `Piwik\Archive` class to query for archive data. If we are allowed to archive in the current
+request/process, and cannot find a recent, usable archive, we generate it, by launching the core archiving process.
+
+### Settings
+
+* browser triggered archiving: this setting determines whether log aggregation is allowed to be launched from browser requests.
```
tsteur: maybe could replace `from browser requests` with `this setting determines whether log aggregation is allowed to be launched on demand when the data is requested by a browser or another client` or something similar?

tsteur: just mentioning it as I always found browser archiving confusing and not clear. It's more like browser archiving == on demand archiving and cli archiving == regular background archiving or so.

docs/4.x/archiving-behavior-specification.md
```diff
@@ -0,0 +1,385 @@
...
+  If disabled, this only happens if a segment that is not set to auto-archive is requested (referred to as a custom segment), or
```
tsteur:
```suggestion
  If browser archiving is disabled, archiving only happens if a segment that is not set to auto-archive is requested (referred to as a custom segment), or
```

docs/4.x/archiving-behavior-specification.md
```diff
@@ -0,0 +1,385 @@
...
+* custom date ranges to pre-process: there is an INI config setting and some user settings that allow users to specify that
+  certain ranges should be pre-archived. The INI setting is `[General] archiving_custom_ranges`. The user setting is the
+  setting that controls the default period to load in Matomo. These ranges will be processed in core:archive if specified.
+
+### Rules
+
+**When archiving is allowed/disallowed for the current process**
+
+If browser archiving is enabled, any HTTP request or CLI command can initiate archiving. Things are more complicated
+if browser archiving is disabled:
```
tsteur, line 51: we might need to mention the configured archiving TTL? Or should it be mentioned in `**Optimizations that skip archiving**`? the ttl applies to browser archiving enabled and disabled. TTL is only looked at I think when date includes today?

also for ranges AFAIK we check the TTL etc.

docs/4.x/archiving-behavior-specification.md
```diff
@@ -0,0 +1,385 @@
...
+
+We know we'd end up with zero visits if we archive, so we can just skip it. The only exceptions are for sites that do
+not use the tracker (so there will always be zero visits in the log table) or are specified via the `'Archiving.getIdSitesToArchiveWhenNoVisits'`
+event. RollUp websites are examples of sites for which this optimization does not apply.
+
+This optimization is used both before launching the archive aggregation logic and in the `core:archive` command
+before we launch individual archive requests. This saves a bit more time since we also don't have to launch an
+archiving command.
+
+### Special Query Parameter Handling
```
tsteur:
```suggestion
### Special URL Query Parameter Handling
```

docs/4.x/archiving-behavior-specification.md
```diff
@@ -0,0 +1,385 @@
...
+Since the range just aggregates day periods, this is still performant, as long as archiving for the days is not launched.
+
+### Launching the core archiving process
+
+The core:archive command launches the core archiving process through new processes, which allows it to achieve concurrency.
+If it's supported by the OS configuration and PHP runtime configuration, these processes are CLI processes spawned through
+`shell_exec()` running the `climulti:request` command. The `climulti:request` command takes a query string and processes it
+like it is an HTTP request.
+
+If it's not supported, then we use CURL multi requests. CURL is not as fast as launching cli processes, so we prefer to use
```
tsteur: I think it's as fast but the problem was that web servers often have request timeouts configured causing the archiving request to be aborted eg after 30 seconds, or after 10 minutes, etc. 

Another bonus was that cli archiving allows us to know how many archivers are currently running (eg across accounts) in the background by executing `ps` etc.

docs/4.x/archiving-behavior-specification.md
```diff
@@ -0,0 +1,385 @@
...
+
+In the core:archive command, we try and process multiple archives at the same time to make the process more performant. The
+number of concurrent processes to run at a time is specified by the `--concurrent-requests-per-website` CLI option.
+
+Archive jobs for different sites are never run concurrently from within the same `core:archive` process. Sites can be
+archived concurrently but this is only possible by running multiple `core:archive` commands.
+
+**Order of Execution**
+
+Some archive invalidations have to be handled before others. Specifically:
```
tsteur, line 247: could mention here the "why". Eg all visits might be executed first because it is viewed by default and looked at most of the time so this data is visible in the UI sooner. Also more recent dates are viewed more often compared to older dates which is why we archive these first

docs/4.x/archiving-behavior-specification.md
```diff
@@ -0,0 +1,385 @@
...
+
+**invalidating custom ranges**
+
+If custom ranges are specified in the `[General] archiving_custom_ranges` INI config setting and/or
+the default period to load in Matomo is configured to be a range, they will be invalidated and
+rearchived during `core:archive`.
+
+**queued invalidations**
+
+Archives that are invalidated from system functions, for example, invalidating past data when a segment or entity changes, are
```
tsteur, line 307: here we could also mention when tracking a request for an older date? Then we remember that archive to be invalidated.

docs/4.x/archiving-behavior-specification.md
```diff
@@ -0,0 +1,385 @@
...
+
+The TTLs are determined by user and config settings. The day period can be specific in the UI in the System Settings page
+and through INI config. The other periods can have custom TTLs in INI config as well, but not in the UI. See above for
+the specific settings.
+
+If no custom TTL is specified for a non-day period, it defaults to the day TTL.
+
+**amount of concurrency to expect**
+
+In general, we can expect that individual core:archive processes will not process multiple sites at the same time. It is
```
tsteur, line 391: I believe technically we can expect multiple sites to be launched at the same time? Even the same site.

Eg if there is only one site, then

* We launch archiver, queue will be eg `$idsites=[1]` and after starting command the queue will be empty
* Next archiver will launch the queue is empty and will be reset to `$idsites=[1]` and the same site will be archived.

Or do we maybe prevent this?

Same could happen from time to time when there are multiple sites... especially if one or few sites take longer to archive while others are faster then the queue will be empty and we reset the list of sites to archive etc.

Or maybe I'm not understanding what is meant by "will not process multiple sites at the same time"

tsteur: actually I see now you're mentioning it re SharedSiteIds

tsteur (commented):
Great read @diosmosis . Left a few comments. What could be also quite useful be another summary with a list of classes and what they do. You are already mentioning this partially during the different sections but I reckon one additional list could be quite useful.

Like 

* `\..\QueueConsumer` - Responsible for ...
* `Piwik\ArchiveProcessor\Loader` - Loads ...
* `... PluginsArchiver` - Does ...
* `... Parameters` - Stores ...

And what could also be useful be a rough/simplified execution flow like (wouldn't need all the details just enough steps to get a rough ID of how it works) 

```
* core:archive command
  * invalidate reports
  * Fetch from archive_invalidations which reports need to be rearchived
  * climulti:request command or a URL request launches to archive data
  * CoreAdminHome.archiveReports API triggered to archive a specific report
     *  Look if there is an archive ID
     * ...
```

docs/4.x/archiving-behavior-specification.md
sgiehl: What happens if core:archive fails to run for a couple of days and is then started again?

diosmosis: Then it should launch archiving for the week which will archive the days in between. This is not ideal but it is basically the same behavior as before, except then we did last52 days.

diosmosis: Btw, this isn't meant to be a review of the archiving code but of this documentation which would hopefully allow you to understand and maintain the system.

sgiehl: Sure. Was just wondering how that is currently solved and if it's worth to mention

diosmosis: I can add "doesn't run for a couple days" as a case here.

docs/4.x/archiving-behavior-specification.md
diosmosis: :+1:

docs/4.x/archiving-behavior-specification.md
diosmosis: :+1:

docs/4.x/archiving-behavior-specification.md
diosmosis: I don't think the TTL is really about when archiving is allowed/disallowed, it's about when is it necessary/unnecessary. I guess it could be Optimizations. It's described below as well.

docs/4.x/archiving-behavior-specification.md
diosmosis: :+1:

docs/4.x/archiving-behavior-specification.md
diosmosis: I remember it saving ~15% per request compared to CURL. Not sure why or whether this was on different hardware (it was one back when piwikpro was around I think so can't remember). Will change it though.

docs/4.x/archiving-behavior-specification.md
diosmosis: Was it also to avoid some unpredictability in archiving?

docs/4.x/archiving-behavior-specification.md
diosmosis: :+1:

docs/4.x/archiving-behavior-specification.md
diosmosis: It mentions that it's possible w/ the current implementation and probably happens regularly if rarely, but it's not the main use case. And that we want things to work when that happens, but are not too concerned with the performance penalty from the edge case.

diosmosis:
@tsteur made changes and replied to some of your comments

docs/4.x/archiving-behavior-specification.md
tsteur: I reckon it's important to be aware of for someone not knowing about it (and it also defines whether archiving will be happening or not in a current process). Either here or optimisations is fine. We could just add a sentence with a link to the section that mentions TTL in more details.

tsteur (commented):
Really nice the added docs around the classes etc 👍

docs/4.x/archiving-behavior-specification.md
```diff
@@ -0,0 +1,508 @@
...
+
+_If an archive's TTL is still valid_
+
+This isn't truly an optimization, but if an archive's TTL is still valid, and it is for an archive where we check
+the TTL (periods that include today and range periods), then we won't launch archiving. See below for more information
+on archive TTLs.
+
+### Special URL Query Parameter Handling
+
+* `trigger`: if set to `archivephp`, let's the core archiving process know it was launched by the `core:archive` command.
```
Findus23:
```suggestion
* `trigger`: if set to `archivephp`, lets the core archiving process know it was launched by the `core:archive` command.
```

docs/4.x/archiving-behavior-specification.md
```diff
@@ -0,0 +1,508 @@
...
+- **Piwik\PluginsArchiver**: This class will loop through every plugin (or requested plugin), create the plugin's `Archiver` instance
+  and invoke it. Based on whether a day or non-day period is being archived, either the `aggregateDayReports()` or `aggregateMultipleReports()`
+  method will be invoked.
+- **Piwik\Plugin\Archiver**: This is the base class for all plugin archivers. Plugins define their archiving logic within one of these
+  classes, which will then be invoked by PluginsArchiver.
+- **Piwik\ArchiveProcessor**: This is a utility class that contains methods to easily query and aggregate metrics and reports for multiple periods.
+  It is mostly used by `aggregateMultipleReports()` implementations to aggregate and insert archive data with a single method call, and by
+  both `aggregateMultipleReports()` and `aggregateDayReports()` to insert archive data.
+- **Piwik\DataAccess\LogAggregator**: This is a utility class that contains some methods to aggregate log data. It is mostly used by
+  `aggregateDayReports()` implementations to aggregate log data and turn the result into `DataTable`s to insert. Many plugin may find
```
Findus23:
```suggestion
  `aggregateDayReports()` implementations to aggregate log data and turn the result into `DataTable`s to insert. Many plugins may find
```

docs/4.x/archiving-behavior-specification.md
```diff
@@ -0,0 +1,508 @@
...
+  a single report.
+
+The entire code path for Core Archiving is as follows:
+
+- An API class constructs an `Archive` instance and queries for archive data.
+- If archiving is not allowed to launch, `Archive` will look through the archive tables directly for data and return whatever it finds.
+- If archiving is allowed, `Archive` will initiate archiving for every site/period combination in the query. It does this by calling the
+  `CoreAdminHome.archiveReports` API method.
+- `CoreAdminHome.archiveReports` creates and invokes the `Loader` class.
+- `Loader` looks for a usable archive. If found, it returns that archive ID. Otherwise it launches the archiving process, first aggregating
```
Findus23:
```suggestion
- `Loader` looks for a usable archive. If found, it returns that archive ID. Otherwise, it launches the archiving process, first aggregating
```

docs/4.x/archiving-behavior-specification.md
```diff
@@ -0,0 +1,508 @@
...
+  core metrics like nb_visits. Then if we can't skip the archive, it invokes `PluginsArchiver`.
+- `PluginsArchiver` loops through every activated plugin and creates the plugin's `Archiver` instance if there is one. It invokes the appropriate
+  method on the `Archiver` instances.
+- `Archiver` instances define `aggregateDayReports()` and `aggregateMultipleReports()` methods. `aggregateDayReports()` methods will
+  generally use a `LogAggregator` instance to aggregate log data and then insert it using an `ArchiveProcessor` instance. `aggregateMultipleReports()`
+  methods will use an aggregation method in `ArchiveProcessor`.
+  - `ArchiveProcessor` aggregation methods will query data using an internal `Archive` instance created using the current period's subperiods
+    (so if the current period is a week, this `Archive` will query for days within that week).
+  - `ArchiveProcessor` will aggregate that data together ans insert it into the archive tables.
```
Findus23:
```suggestion
  - `ArchiveProcessor` will aggregate that data together and insert it into the archive tables.
```

docs/4.x/archiving-behavior-specification.md
```diff
@@ -0,0 +1,508 @@
...
+
+* **Piwik\Plugins\CoreConsole\Commands\CoreArchiver**: The `Command` class for the `core:archive` command. This will create and use a `CronArchive`
+  instance. This is the main code path from which `CronArchive` is invoked, the only other being the `CoreAdminHome.runCronArchiving` API method.
+* **Piwik\CronArchive**: Encapsulates the entire cron archiving process. This class will loop over every requested site (or work on an existing
+  shared queue of site IDs), invalidate data tha needs to be invalidated for it, and process archives for every invalidation for the site.
+* **Piwik\CronArchive\QueueConsumer**: This class is used by CronArchive to get the next invalidated archives to process. It queries the
+  archive_invalidations table and returns a batch of archive invalidations to process. CronArchive will then launch API requests to
+  `CoreAdminHome.archiveReports` for those invalidations.
+* **Piwik\CliMulti**: This is a utility class used by CronArchive to initiate multiple API requests in parallel. It will try to execute the requests
+  from the CLI using the `climulti:request` command if it can. Otherwise it falls back to initiating curl requests.
```
Findus23:
```suggestion
  from the CLI using the `climulti:request` command if it can. Otherwise, it falls back to initiating curl requests.
```

docs/4.x/archiving-behavior-specification.md
```diff
@@ -0,0 +1,508 @@
...
+* **Piwik\Archive\ArchiveInvalidator**: This is a core service class that is not strictly part of the core:archive system, but is very important
+  to it. It contains the logic to invalidate archive data, and `CronArchive` uses it (indirectly) to invalidate archives before it starts processing
+  for a single site.
+
+The entire code path for core:archive execution is:
+
+- The `CoreArchiver` command class is invoked. It processes user input and constructs a `CronArchive` instance. Options are set on the instance and
+  it is invoked.
+- `CronArchive` uses `FixedSiteIds` or `SharedSiteIds` to start pulling sites to archive. For every site in that list:
+  - `CronArchive` invalidates any data that needs to be invalidate for the site. This can be from visits tracked in the past, entities like segments that were
```
Findus23:
```suggestion
  - `CronArchive` invalidates any data that needs to be invalidated for the site. This can be from visits tracked in the past, entities like segments that were
```

docs/4.x/archiving-behavior-specification.md
```diff
@@ -0,0 +1,508 @@
...
+    created or updated or special periods specified in config or settings. The invalidations are inserted into the `archive_invalidations` table.
+  - `CronArchive` then uses `QueueConsumer` to pull invalidations to process from the `archive_invalidations` table which pulls them in the correct execution
+    order.
+  - `CronArchive` takes the batch of invalidations from `QueueConsumer` and uses `CliMulti` to launch multiple archiving processes via API method.
+  - `CronArchive` looks at the results of those API calls and determines whether the request was successful. If so, the invalidation is removed. Otherwise
```
Findus23:
```suggestion
  - `CronArchive` looks at the results of those API calls and determines whether the request was successful. If so, the invalidation is removed. Otherwise,
```

docs/4.x/archiving-behavior-specification.md
```diff
@@ -0,0 +1,508 @@
...
+  - This continues until `QueueConsumer` reports no invalidations to process.
+- We pull a new site and process it. This continues until there are no sites to process.
+
+### Range Archiving vs All Others
+
+By default when using core:archive, day, week, month & year periods are not processed in the browser. Range periods
+on the other hand are still processed from web requests. This is because we cannot pre-archive every possible permutation
+of dates that users can specify.
+
+Instead when a range period is no longer considered up-to-date (it's TTL expires), we launch archiving for the range.
```
Findus23:
```suggestion
Instead when a range period is no longer considered up-to-date (its TTL expires), we launch archiving for the range.
```

Findus23 (changes requested):
just a few quick typo fixes

diosmosis (commented):
* [x] TODO: should add a more detailed section on archive purging as well, along w/ potential issues there for support purposes

docs/4.x/archiving-behavior-specification.md
diosmosis: :+1:

diosmosis:
fyi @sgiehl, @tsteur ready to review or merge

tsteur (approved):
All good from my perspective. Great to have this doc @diosmosis 🎉 💯

PR #415, opened 2021-01-18
Add quick blurb about translateColumnNames parameter.

### LINKED ISSUE #411: Document the URL parameter 'translateColumnNames=1' in API reference
we had a few support requests where people asked about this parameter, it would be valuable to document it in https://developer.matomo.org/api-reference/reporting-api

diosmosis:
Fixes #411


diosmosis:
@sgiehl can you give a quick review?

docs/4.x/reporting-api.md
```diff
@@ -113,6 +113,8 @@ There are also generic filters you can choose to apply on all APIs that return w
 *   **disable\_generic\_filters**; if set to 1, all the generic filters above will not be applied. This can be useful to disable the filters above which are otherwise applied with default values. Mostly used internally or when developing plugins for Piwik.
 *   **disable\_queued\_filters**; if set to 1, all the filters that are mostly presentation filters (replace a column name, apply callbacks on the column to add new information such as the browser icon URL, etc.) will not be applied. Mostly used internally or when developing plugins for Piwik.
 
+*   **translateColumnNames**; if set to 1, column names in report output will be translated to the current user's language (the language is set in the user's personal settings).
```
sgiehl: IIRC this param is only available for CSV, RSS and HTML export. Not sure if it has an impact on XML at all... The language can actually be set using `language` parameter, but should default to the user's language I guess...

diosmosis: changed

sgiehl (approved):

PR #418, opened 2021-01-29
Remove JS require documentation

tsteur:
AFAIK we're not doing this anymore already for a long time and I reckon it's rather confusing and could be removed. We're usually always using now angularjs anyway and we'll migrate things to angular


sgiehl (approved):
I guess we can't yet use angular for everything. E.g. when extending the datatable javascript, which is not angular yet... But might be fine to remove that already.

PR #419, opened 2021-01-29
Mention broadcast and ajaxhelper is deprecated

tsteur:


sgiehl (approved):

docs/4.x/working-with-piwiks-ui.md
```diff
@@ -167,6 +167,8 @@ Piwik Core defines many classes that should be reused by new plugins and contrib
 
 ### Broadcast
 
+This class is considered deprecated and only exists for historical reasons as some components are still using it. You should usually instead use angularjs components like [$location](https://docs.angularjs.org/api/ng/service/$location).
```
diosmosis: `$location` actually doesn't always work the way we want. For example, it does not handle array parameters right (IIRC). I don't think we can deprecate it.

diosmosis (changes requested):
Not sure we can deprecate broadcast, it's handling of URLs is actually more accurate than using `$location`, though it's been a while so I can't remember everything that doesn't work with `$location`.

docs/4.x/working-with-piwiks-ui.md
tsteur: It's more like discouraged to be used and angularjs is preferred whenever any possible. I'll make it more clear 👍

PR #423, opened 2021-02-16
update docs for run the tests at first time

flamisz:
Small fixes to make running the tests first time easier.

### DIFF
--- a/docs/4.x/tests-php.md
+++ b/docs/4.x/tests-php.md
@@ -28,7 +28,17 @@ If your development Matomo is setup in a sub-directory for example at http://loc
 ```
 [tests]
 request_uri = "/dev/matomo"
+```
 
+Before you run the tests (at least the first time, but you can rerun it any time), run this command to migrate the test database. _Please note:_ to run this command you **have to** add the `request_uri` settings (it can be an empty string) into the `config/config.ini.php`. After the command finished, you can delete it.
+
+```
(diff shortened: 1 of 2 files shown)


docs/4.x/tests-php.md
```diff
@@ -28,8 +28,18 @@ If your development Matomo is setup in a sub-directory for example at http://loc
 ```
 [tests]
 request_uri = "/dev/matomo"
+```
+
+Before you run the tests (at least the first time, but you can rerun it any time), run this command to migrate the test database. _Please note:_ to run this command you **have to** add the `request_uri` settings (it can be an empty string) into the `config/config.ini.php`. After the command finished, you can delete it.
```
Findus23: I think (but don't know) that there are some tests that also need the request_uri set

flamisz: Yes, you're right, the `system` test needs it and doesn't even like the empty string, but happy with the `"/"`. 

I'll add to this copy a warning about some tests may need this config.

sgiehl: Wouldn't it be better to simply adjust the part above, and mention that `request_uri` needs to be configured for running tests. It currently only mentions that it is needed when installed in a subdirectory. But actually it should always be set.

tsteur:
Cheers @flamisz 👍

PR #424, opened 2021-02-17
Add additional documentation for when creating a PR.

tsteur:
fyi @flamisz might be interesting for you.


tsteur:
BTW do we still need the `WIP Work In Progress` label since we have github drafts?

PR #428, opened 2021-02-23
Document switching branches and keeping things up to date

tsteur:
@diosmosis @sgiehl be great to review and add info to the docs where possible.

### DIFF
--- a/docs/contributing-to-piwik-core.md
+++ b/docs/contributing-to-piwik-core.md
@@ -202,6 +202,15 @@ update matomo_option set option_value ='{new_version_number_you_want_to_set}' wh
 
 Note: Matomo updates are designed to always be executed multiple times. If the update already happened, for example a column was already added, then Matomo will "detect" this and not error. 
 
+### Downgrading Matomo updates (not possible)
+
+Sometimes you may need to run some code on a previous version of Matomo, for example, making a change to Matomo 3. There's no
+current way to downgrade a database (undo an update) so in this case you'll have to use an upgraded version of Matomo
+with the older code.
+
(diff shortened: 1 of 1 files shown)


docs/contributing-to-piwik-core.md
```diff
@@ -160,6 +160,49 @@ Here are best practises we aim to follow when creating, reviewing and merging pu
 * Read the article: [Pull Requests: How to Get and Give Good Feedback](https://www.kickstarter.com/backing-and-hacking/pull-requests-how-to-get-and-give-good-feedback)
 * When a pull request has many small commits, it is recommended to [Squash commits](http://eli.thegreenplace.net/2014/02/19/squashing-github-pull-requests-into-a-single-commit) so that there is only one (bigger) commit.
 
+## Switching between branches
+
+### Keeping dependencies up to date
+
+Every time you change the branch you may need to run composer install (and maybe also npm install). You can automate this task by following these steps:
```
diosmosis, line 167: It's pretty rare that this needs to be done, isn't it? We only update when needed to (security fixes or major version), so this might make switching branches slower (I don't use it so I'm not sure).

tsteur: It's rare it needs to actually install a newer version but when it's needed then it can be hard to find out why it's not working or harder to reproduce things since you're not testing correct version. And when there's nothing to update then it executes fairly fast as it only executes when the file actually changes (so only when needed).

diosmosis:
Added a quick section on using an older matomo version w/ a newer database version. Will keep this open in case @sgiehl wants to take a look

sgiehl (approved):
LGTM

PR #439, opened 2021-03-05
Add note about disabling marketplace plugin for faster loading of admin pages locally.

diosmosis:
FYI @tsteur


tsteur:
Good one, I do that too 👍

tsteur:
fyi @flamisz might be useful to know

PR #440, opened 2021-03-05
add description of how to use controller methods to return json results

diosmosis:
FYI @tsteur

### DIFF
--- a/docs/4.x/controllers.md
+++ b/docs/4.x/controllers.md
@@ -134,7 +134,7 @@ public function myAdminTask()
 
 In general, most functionality in Matomo controllers will render some HTML and return it. Sometimes, however you will have some logic that's
 meant to return JSON, but is something you don't want exposed in an API method (as in, callable outside of a session). In this case, we create
-controller methods that return JSON.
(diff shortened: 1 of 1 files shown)


docs/4.x/controllers.md
```diff
@@ -130,6 +130,32 @@ public function myAdminTask()
 }
 ```
 
+### Using Controller Methods as API Methods
+
+In general, most functionality in Matomo controllers will render some HTML and return it. Sometimes, however you will have some logic that's
+meant to return JSON, but is something you don't want exposed in an API method (as in, callable outside of a session). In this case, we create
+controller methods that return JSON.
```
tsteur: Doc is 👍  could we maybe add at the end of this sentence maybe something like `Most of the time adding an API method instead of a controller method should be considered.`. You mention it already in this block but I'm thinking we might still want to make clear in most cases an API is likely the better solution. Or maybe instead of adding this we could explain with an example when the controller would be used instead of the API? I guess it's mostly only the case it the controller needs to access the session?  Maybe that's what you meant with `callable outside of a session`?

diosmosis: This was in regards to the Feedback controller methods that set the next reminder date. These methods don't output HTML or do any rendering, but of course there's no reason to put them in the API because why would anyone need to call them? I'm not sure how to better describe these outlier use cases.

PR #441, opened 2021-03-06
add note about MDN

diosmosis:
personally I find it more useful than caniuse

### DIFF
--- a/docs/core-faqs.md
+++ b/docs/core-faqs.md
@@ -8,6 +8,8 @@ title: FAQs
 
 Go to the [Can I use website](http://caniuse.com/) and enter the name of the feature you want to use. It will tell you which browsers support this specific feature. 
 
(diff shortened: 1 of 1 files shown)


Findus23:
Just FYI: caniuse now also uses MDN data, so you can now also find compatibility data for very specific funktions (e.g. https://caniuse.com/mdn-api_domtokenlist_contains) there.

tsteur:
👍  LGTM. BTW @diosmosis curious where the information is more granular or what you mean there? Just FYI the reason I use ICanUse is that you can easily search for anything and see the information directly without needing to scroll or click around. And you can filter and show different visualisations when needed etc.

diosmosis:
@tsteur it's been quite a while since i've had to look something up, but generally if I want to know if a specific function is usable on a specific browser (eg, Array.find), MDN makes it easier since the docs for that specific method will say so (the same for node attributes, language features, etc.). W/ caniuse (in the past at least), a search would lead to "Array functions" or something generic, and I wouldn't necessarily know if the specific thing I was looking for was included.

tsteur:
Thanks @diosmosis

PR #453, opened 2021-03-16
Update API Core Doc

### LINKED ISSUE #8126: Our work process when changing or removing a public API from the Piwik platform
In this issue we describe our process when we want to change or remove a public API. The goal is to put this knowledge in a guide on developer.piwik.org so it is easy to spread this knowledge to all team members including new ones in the future.
### Our Backwards Compatibility Promise

All popular software platforms have a process to ensure Backward Compatibility (BC) is kept between `Minor` and `Patch` releases (see [Semantic Versioning 2.0.0](http://semver.org/)). when BC is kept, it means users can be confident to upgrade to a newer version (Minor or Patch release) that their platform will still work (including any installed third party plugins.). For example Symfony have a very advanced BC guide: [Our Backwards Compatibility Promise ](http://symfony.com/doc/current/contributing/code/bc.html).
### Process to keep BC, and deprecate APIs

Here are the notes how our current process work:

Given:
- Piwik uses Semver (#4796)
- Piwik is an open platform that developers can trust,
- term API is used to refer to many types of APIs (see #8125 for details),

Then:
- We do not break BC in a `Minor` or `Patch` release
  - we may break BC in a Minor or Patch release, in rare cases only when BC break is necessary for Security reasons or for Performance reasons
- When we need to change an API, or remove an API, before removing or changing the API, we deprecate it:
  - this can usually be done by adding `@deprecated` tag in the API, event name, etc.
  - we announce the deprecation in the Developer Changelog at least 3-6 months early
- As defined in SemVer, Major releases are our only chance to remove our deprecated code.
  - When we release a `Major` version (eg. [Piwik 3.0.0](https://github.com/piwik/piwik/milestones/3.0.0)) then we are will remove all `@deprecated` code and therefore break BC.  
  - we announce the details of code removed in the developer changelog #8127
  - ideally we also document to developers how they can convert their code to the new way

### LINKED ISSUE #8127: Developer Changelog and documenting changes in the Piwik Platform
The goal of this issue is to list all the changes that we aim to document in our Developer Changelog ([/CHANGELOG.md](https://github.com/piwik/piwik/blob/master/CHANGELOG.md)). We need to move this knowledge to a guide in developer.piwik.org so that it's shared across team members, and clear to the developer community.
### Context

Where we are today:
- We create the Piwik platform and offer developers a set of public APIs #8125 
- We maintain backward compatibility for those APIs. 
- When changing or removing public APIs, we must first deprecate code following this process: #8126
- We keep track of all platform changes relevant to developers in our [Developer changelog](http://developer.piwik.org/changelog)
### What to write in developer changelog

As a Piwik product team engineer, we keep the developer changelog up to date by noting any:
- change to a Reporting API method (eg. new API method added, deprecated, removed)
- change to Reporting API output (eg. a new field in an API response)
- change to Reporting API parameter (eg. parameter added, deprecated, removed)
- change to Tracking API parameter (eg. parameter added, deprecated, removed)
- change to Piwik JavaScript Tracker feature (eg. new feature, or removed feature)
- since PHP 8 argument names of public API's are also considered API
- new console command 
- new parameter for a console command
- new developer guide
- update to a third party library  
- any other relevant internal change that may interest developers
### Developer changelog structure

These changes are organised in the developer changelog in the following categories:
- Breaking Changes
- Deprecations
- New features
- New APIs
- New commands
- New guides
- Library updates
- Internal change

(you will find the template at the bottom of `CHANGELOG.md` file)

tsteur:
The page https://developer.matomo.org/guides/apis contains outdated information and pretty much all is now documented in https://developer.matomo.org/guides/expose-api-methods anyway so removing this content and added more relevant content for the core team.

refs https://github.com/matomo-org/matomo/issues/8125
fix https://github.com/matomo-org/matomo/issues/8126
fix https://github.com/matomo-org/matomo/issues/8127


tsteur:
fyi merging this and another PR already as I will be moving few more things around. Let me know if there are any changes needed and I'll be happy  to make them.

PR #460, opened 2021-03-26
Mention xdebug_config wont work with xdebug 3

tsteur:
tested this today for quite a while and tried to get it working but it doesn't work with Xdebug 3. See https://xdebug.org/docs/all_settings only some settings support `XDEBUG_CONFIG` and neither `start_with_request` nor the `mode` option supports it.


tsteur:
Note: I tested running some tests with `xdebug.mode=debug` and `xdebug=off` and the overhead of xdebug is not that much. Eg 1.15minutes instead of 1.25minutes.

PR #463, opened 2021-03-29
Document duplication of translation keys

tsteur:
refs https://github.com/matomo-org/matomo/pull/17380


Findus23:
I just noticed that the bottom of the page directly contradicts this advice:

>  Reduce redundancy in your translated text. If the same text appears in multiple translated text entries, try to move the translated text out by using sprintf parameters. For example, if you have text entries like:
> 
> "You cannot use commas." and "You cannot use underscores."
> 
> Try to split them up into something like:
> 
> "You cannot use %s." "commas" and "underscores".
> 
> This guideline is more important for contributions to Matomo Core than for new plugins.

I guess we should just remove that part

PR #464, opened 2021-03-29
Update caching-in-matomo.md

flamisz:


Findus23:
Small ad (in case it helps you): This is why I created https://plugins.matomo.org/DevelopmentToogle which just adds a tiny icon in the top bar to toggle development mode (as disabling it speeds things up a lot when you are just trying out non-development things).

PR #465, opened 2021-03-31
Create documentation for Updates

diosmosis:


flamisz:
Very useful, thanks. 
I have one question, that maybe can be included in this doc:
what happens when we upgrade Matomo skipping versions? Like from 3.1.0 to 4.1.2? How does the updater know what is the order of the updates and does it run 1-by-1, etc?

diosmosis:
@flamisz I added a quick update here: https://github.com/matomo-org/developer-documentation/pull/465/files#diff-5b35eacc27e9998b990dddd6abe0c090feab5f674d9f7eca330ae9bd992ef153

It doesn't go over automatic dimension updates. I'd have to look through the code to write that, so that will have to wait.

docs/4.x/updates-aka-migrations.md
```diff
@@ -0,0 +1,102 @@
...
+   logic you need.
+
+## How Updates Are Run
+
+On every request Matomo will check if the current version of Matomo (stored in the `version_core` option) is less than the version hardcoded in Version.php.
+If this is true, then the code was updated, but not the database. This triggers the database update workflow.
+
+The update workflow looks for every update file in the `Updates` folder that has a version greater than the current version and less than or equal to the version in Version.php (using `version_compare()`). It runs each in order, one after the other. Once done, the `version_core` option is updated.
+
+The updater will also automatically make changes to tracking dimensions if their definition changes.
```
tsteur, line 92: Can we maybe move the content from https://developer.matomo.org/guides/contributing-to-piwik-core#dealing-with-matomo-updates in here somewhere too? And then keep the title in https://developer.matomo.org/guides/contributing-to-piwik-core#dealing-with-matomo-updates but change the paragraph to only link to this guide similar to 
image

docs/4.x/updates-aka-migrations.md
```diff
@@ -0,0 +1,102 @@
...
+    public function doUpdate(Updater $updater)
+    {
+        $updater->executeMigrations(__FILE__, $this->getMigrations($updater));
+    }
+}
+```
+
+The `getMigrations()` method returns explicit step by step changes to make to the system. Separating them like
+this allows us to display the list of changes that will be made to the user in case the user would like to do
+them manually, one at a time (useful for large instances).
```
tsteur, line 55: can we maybe link to https://developer.matomo.org/api-reference/Piwik/Updater/Migration/Factory and mention that there are various methods exist to make common changes to a schema like adding/removing a column or table, activating or deactivating a column, etc. see 

https://developer.matomo.org/api-reference/Piwik/Updater/Migration/Db/Factory
https://developer.matomo.org/api-reference/Piwik/Updater/Migration/Plugin/Factory
https://developer.matomo.org/api-reference/Piwik/Updater/Migration/Config/Factory

I reckon it could be short for now but eventually we could then also add examples.

tsteur (changes requested):
@diosmosis great doc 👍 

Can we maybe also add additionally that before merging a PR with an update, we need to check if there have been releases in between and if so, update the version for the specific update?

docs/4.x/updates-aka-migrations.md
diosmosis: Those docs are more about dealing with the need to update while developing Matomo, while this guide is about how they are written and how they generally work. I can't think of a good place to put those two settings in this guide, so I just linked to it in contributing-to-piwik-core

docs/4.x/updates-aka-migrations.md
diosmosis: :+1:

docs/4.x/updates-aka-migrations.md
tsteur: 👍

PR #467, opened 2021-03-31
Add troubleshooting section for system test w/ answer for "Only generated N API calls..." error

diosmosis:


docs/4.x/tests-system.md
```diff
@@ -98,3 +98,24 @@ You can create your own fixture as well, just extend `Piwik\Tests\Framework\Fixt
...
+...
+```
+
+This can have the following possible causes:
+
+* You're trying to test API methods that do not start with `get`. System tests are designed only to be able to test
+  `get` API methods.
+* You're missing a required parameter in a specific API call or it's being set to `null` or equivalent. To determine
+  what the root issue is here, you'll need to inspect the variable `$nameVariable` in this if statement:
+  [https://github.com/matomo-org/matomo/blob/4.x-dev/core/API/DocumentationGenerator.php#L40-L41](https://github.com/matomo-org/matomo/blob/4.x-dev/core/API/DocumentationGenerator.php#L40-L41)
```
tsteur: @diosmosis could maybe like to a specific version like https://github.com/matomo-org/matomo/blob/4.2.0/core/API/DocumentationGenerator.php#L40-L41 as the lines might change over time?

diosmosis: :+1:

PR #473, opened 2021-04-07
also unset Custom Dimensions when using SPAs

Findus23:
I have not tested this, but I assume this is also needed.

Maybe Matomo could also provide a function to delete all Custom Dimensions in one function call.

### DIFF
--- a/docs/4.x/spa-tracking.md
+++ b/docs/4.x/spa-tracking.md
@@ -37,6 +37,15 @@ _paq.push(['trackPageView']);
 
 Note: we recommend you use [Custom Dimensions](https://matomo.org/docs/custom-dimensions/) instead of Custom Variables as they will be deprecated in the future.
 
+### Resetting previously set custom dimensions
+
+Similar to Custom Variables, you also need to unset [Custom Dimensions](https://matomo.org/docs/custom-dimensions/) when changing the page as they would otherwise be tracked again.
+
+```javascript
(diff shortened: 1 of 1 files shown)


docs/4.x/spa-tracking.md
```diff
@@ -37,6 +37,15 @@ _paq.push(['trackPageView']);
 
 Note: we recommend you use [Custom Dimensions](https://matomo.org/docs/custom-dimensions/) instead of Custom Variables as they will be deprecated in the future.
 
+### Resetting previously set custom dimensions
+
+Similar to Custom Variables, you also need to unset [Custom Dimensions](https://matomo.org/docs/custom-dimensions/) when changing the page as they would otherwise be tracked again.
```
tsteur, line 42: fyi @Findus23 looking at code they would also be sent with other actions (outlink, download, event) so wonder if they would need to be reset best after a `trackPageView` if it is a dimension in scope action. Like 

```
_paq.push(['trackPageView']);
_paq.push(['deleteCustomDimension', 1]);
```

Of course depends on the use case.

Findus23: That indeed depends on the use-case.

But on a "normal" page the custom dimension is also set when outlinks, etc. are tracked later on.
So I guess to stay as similar as possible to the "normal" use-case, I think deleting them on a route change is probably the best.

PR #482, opened 2021-05-25
Add quick docs on sending emails in Matomo.

diosmosis:


docs/4.x/sending-emails.md
```diff
@@ -0,0 +1,101 @@
...
+All mail sending functionality in Matomo goes through the `Mail` class, which in turn uses the PHPMailer library.
+
+Using the class is straightforward:
+
+```php
+$config = \Piwik\Config::getInstance();
+
+$mail = new \Piwik\Mail();
+$mail->addTo($targetEmail, 'Matomo User');
+$mail->setFrom($config->General['noreply_email_address'], $config->General['noreply_email_name']);
```
tsteur: not  sure would we call `setDefaultFromPiwik` here instead? (could maybe also add `setDefaultFromMatomo` method proxying to `setDefaultFromPiwik`)

diosmosis: I just took this from the test-email command. Will change it.

tsteur:
👍  great

PR #484, opened 2021-06-01
Add strategies doc for debugging tracking code on live websites.

diosmosis:


docs/4.x/debugging-strategies.md
```diff
@@ -0,0 +1,52 @@
+---
+category: DevelopInDepth
+---
+
+This document lists strategies that can be useful in debugging the many different parts of Matomo while developing on the platform.
+You might find these useful if you are contributing to core or are writing your own plugin.
+
+# Debugging Tracking Code
```
tsteur: could we name this `Debugging the JS tracker`? because there is already debugging the tracker guide https://developer.matomo.org/api-reference/tracking-api#debugging-the-tracker

tsteur (commented):
@diosmosis left a comment.
 
Generally wondering if we should put this section into https://developer.matomo.org/guides/jstracker-core

and then link to it from https://developer.matomo.org/guides/debugging-core similar as we did with https://developer.matomo.org/guides/debugging-core#debugging-the-tracker

flamisz (approved):
It's very useful, I like it. 👍 
I didn't know about that `--ignore-minified` trick, I usually just used the link to the `js/piwik.js` instead of `piwik.js` in the source of the javascript.

docs/4.x/debugging-strategies.md
diosmosis: @tsteur modified

tsteur:
👍  thanks for that

PR #486, opened 2021-06-22
Document updating screenshots on the website when needed

tsteur:
not sure how to best describe it. Something we probably didn't do as much in the past but should when the UI changes quite a bit.

### DIFF
--- a/docs/pull-request-reviews.md
+++ b/docs/pull-request-reviews.md
@@ -165,6 +165,8 @@ but for smaller changes, developers might be expected to make those themselves.
 
 We should also think about whether new faqs should be created or if we need to modify existing ones.
 
+If a screenshot changes significantly, then we should also update screenshots on our website. If only a wording changes, then usually it's not needed to update the screenshot. There are no hard rules when or when not to update a screenshot. If basically depends if the screenshot is still clear in the context of the content. 
+
 ### Reviewing External Contributors' PRs
 
 External contributor pull requests should be reviewed in the same way as PRs from core developers, EXCEPT:


docs/pull-request-reviews.md
```diff
@@ -165,6 +165,8 @@ but for smaller changes, developers might be expected to make those themselves.
 
 We should also think about whether new faqs should be created or if we need to modify existing ones.
 
+If a screenshot changes significantly, then we should also update screenshots on our website. If only a wording changes, then usually it's not needed to update the screenshot. There are no hard rules when or when not to update a screenshot. If basically depends if the screenshot is still clear in the context of the content.
```
sgiehl, line 168: Isn't that something we should only do after a release? It might confuse people if we have newer screenshots on the website and they don't match the latest release maybe.

tsteur: We would likely forget just like updating the documentation. Considering we eventually want to release regularly (every 4 weeks) and updating screenshots isn't happening that often it might be just easier/simpler to do it right away. If the team prefers to create a new issue to do it just before the release this could work too though (just a bit of overhead maybe).

sgiehl: To be honest I don't know which parts of Matomo actually have a screenshot somewhere on matomo.org. And it kind of would be a waste of time looking through the website for screenshots that might need an update. I guess the most useful thing would be to have that automated. We could use headless chrome to automatically fetch screenshots from specific demo urls once a month (or less/more often) and provide an automatic PR to update them on the website repo. Might be a bit more work to initially make that work, but it would save a lot time later.

tsteur: @sgiehl we discussed this actually recently also with @justinvelluppillai but it's not that simple and it's a lot lot more work. For example some screenshots showing reports requires very specific data in the reports, requires tracking a lot of visits etc. Also regarding sizes etc and what exactly needs to be shown in can get complicated quite quickly. It's mostly searching for related user guides on the website just like the way we search for whether any other documentation or FAQ or guide needs to be updated. Nothing actually changes there except for not only looking for text but also looking for images when checking if we need to update any documentation.

diosmosis: A random thought: for specific data, the cloud demo could be used (there might be some data missing there that might need to be fixed). Actually... we could also use iframes to embed read only versions of the demo. I'm not sure if that would be less work though.

tsteur: @diosmosis would need to see but that may be sometimes bit slow to load compared to a small image (Matomo is few MB but of course cached after loading once). What could be pretty awesome though could be maybe first loading an image in the future with some kind of play icon and when clicking on it replacing it with a frame so visitors can interact with it :)

diosmosis: :+1: that's definitely a cool idea!

PR #488, opened 2021-06-24
recommend specifying a base branch when checking out a new branch

tsteur:


sgiehl (approved):

Findus23:
Thanks for the trick, I didn't know it before. I would always checkout 4.x-dev before creating the branch, but this would do it in one step.

PR #494, opened 2021-07-21
Mentioning some examples for how php API might break

tsteur:
That's not for events though which is different and needs to be added too eventually


diosmosis:
THere's a place in the reviewing prs doc that might be a more relevant place

tsteur:
changed it @diosmosis 👍

PR #496, opened 2021-07-28
Proposal to start documenting new INI settings in developer changelog

mattab:
I wasn't sure if it's something we want to do more consistently, i've noticed we've started to document them recently, and i find it helpful. Although not 100% relevant to platform developers, it is still better than not having it maybe?

### DIFF
--- a/docs/4.x/apis.md
+++ b/docs/4.x/apis.md
@@ -82,6 +82,7 @@ When we are adding a new API or when we are breaking or deprecating an existing 
 * change to Piwik JavaScript Tracker feature (eg. new feature, or removed feature)
 * since PHP 8 argument names of public API's are also considered API
 * new console command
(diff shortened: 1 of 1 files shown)


tsteur:
@mattab it be great to keep the changes relevant for developers and would rather not have it there

justinvelluppillai:
Has our position changed on this? Suggest we either merge or close this PR, I'm happy either way. @mattab 's suggestion is consistent with how we use the changelog, we do have some releases where we specifically mentioned new ini settings.

sgiehl:
Imho it makes sense to document those new settings in the changelog. Even though most settings won't be needed for a developer, there might be cases where someone should/could use such a new setting. And it's always good for a developer to know which settings are available and can be used if needed.

PR #497, opened 2021-07-28
Add an FAQ for tests

tsteur:
Might not be super detailed yet but I think it's a lot better than having nothing and gives us ability to add more over time and clarify things.

### DIFF
--- a/docs/4.x/tests-in-depth-faq.md
+++ b/docs/4.x/tests-in-depth-faq.md
@@ -9,7 +9,7 @@ category: DevelopInDepth
 
 You can use the [configuration object](/guides/piwiks-ini-configuration) as usual and set any specific ini setting before executing the test.
 
-You don't need to reset the ini after each test as this is done automatically for you.
+In integration tests you don't need to reset the config ini after each test as this is done automatically. In system tests the ini change is kept for all tests in the same class unless you manually undo the ini change.
 
 You can find an example in the [BlobReportLimitingTest](https://github.com/matomo-org/matomo/blob/4.4.0/tests/PHPUnit/System/BlobReportLimitingTest.php#L194-L216). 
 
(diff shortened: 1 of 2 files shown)


docs/4.x/tests-in-depth-faq.md
```diff
@@ -0,0 +1,144 @@
...
+---
+# Tests FAQ 
+
+## How do I modify the config ini to test certain conditions?
+
+### Integration and system PHP tests
+
+You can use the [configuration object](/guides/piwiks-ini-configuration) as usual and set any specific ini setting before executing the test.
+
+You don't need to reset the ini after each test as this is done automatically for you.
```
tsteur: not sure if this is also true for system tests? whether we reset after each test method?

docs/4.x/tests-in-depth-faq.md
```diff
@@ -0,0 +1,144 @@
...
+
+You can use either `/config/environment/test.php` or a plugin specific `config/test.php` say `plugins/CoreHome/config/test.php` to overwrite any dependency in Matomo during tests.
+
+For examples see [config/test.php](https://github.com/matomo-org/matomo/blob/4.4.0/config/environment/test.php) and [Tour/test.php](https://github.com/matomo-org/matomo/blob/4.4.0/plugins/Tour/config/test.php).
+
+Changes here are usually applied to all tests.
+
+### Integration and system PHP tests
+
+If you want to apply changes only for a specific test then you can overwrite the `provideContainerConfigBeforeClass` or the `provideContainerConfig` method to change the injected dependencies if neeeded.
```
tsteur: not 100% sure when to use which of the two methods. When would we use one or the other?

docs/4.x/tests-in-depth-faq.md
diosmosis: not true for system tests, just integration

docs/4.x/tests-in-depth-faq.md
diosmosis: provideContainerConfig for IntegrationTestCase and UnitTestCase classes, provideContainerConfigBeforeClass for SystemTestCase, since system tests don't do the reset logic as mentioned above. The override will be for an entire test case though, not just a single test.

docs/4.x/tests-in-depth-faq.md
tsteur: 👍  thanks for this

docs/4.x/tests-in-depth-faq.md
tsteur: great, thanks 👍

PR #498, opened 2021-07-29
Add documentation for system checks

tsteur:
refs https://github.com/matomo-org/matomo/pull/17832

It's not perfect but at least provides a hint of how system checks can be added etc.

### DIFF
--- a/docs/4.x/system-check.md
+++ b/docs/4.x/system-check.md
@@ -44,18 +44,18 @@ First you need to decide if your system check is of type required, optional or i
 
 ```php
 return [
-    'diagnostics.required' => array(
+    'diagnostics.required' => DI\add(array(
         // adds a required system check
         DI\get('Piwik\Plugins\MyPluginName\Diagnostic\JsDirectoryWritePermissionCheck'),
-    ),
-    'diagnostics.optional' => array(
+    )),
+    'diagnostics.optional' => DI\add(array(
         // adds an optional system check
         DI\get('Piwik\Plugins\MyPluginName\Diagnostic\JsDirectoryWritePermissionCheck'),
-    ),
-    'diagnostics.informational' => array(
+    )),
+    'diagnostics.informational' => DI\add(array(
         // adds an informational system check
         DI\get('Piwik\Plugins\MyPluginName\Diagnostic\JsDirectoryWritePermissionCheck'),
-    )
+    ))
 ];
 ```
 
--- a/docs/4.x/system-check.md
+++ b/docs/4.x/system-check.md
@@ -92,7 +92,7 @@ public function execute()
 Each check can have one of four different statuses:
 
 * **OK** (`DiagnosticResult::STATUS_OK`) - Everything is fine and the system passes the check.
-* **Error** (`DiagnosticResult::STATUS_ERROR`) - There is an error which does need to be fixed. Typically used for a required check.
+* **Error** (`DiagnosticResult::STATUS_ERROR`) - There is an error which needs to be fixed. Typically used for a required check.
 * **Warning** (`DiagnosticResult::STATUS_WARNING`) - There is a warning and it is optional to fix this. Typically used for an optional check.
 * **Informational** (`DiagnosticResult::STATUS_INFORMATIONAL`) - Only provides information. Typically used for an informational check.
 
--- a/docs/4.x/system-check.md
+++ b/docs/4.x/system-check.md
@@ -1,0 +1,137 @@
+---
+category: Develop
+title: System check
+---
+# Enriching the system check
+
+Some plugins may only work when certain extra conditions are met, like having an unusual PHP extension installed or needing certain write permission on the file system. In these cases it is recommended to add new checks to the Matomo diagnostics system check which you can find by going to `Matomo Administration -> Diagnostic -> System Check`. 
+
+## Types of system checks 
+
+Matomo typically differentiates between three kind of checks:
+
+### A required check
+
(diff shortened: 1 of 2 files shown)


docs/4.x/system-check.md
```diff
@@ -0,0 +1,137 @@
...
+    ),
+    'diagnostics.optional' => array(
+        // adds an optional system check
+        DI\get('Piwik\Plugins\MyPluginName\Diagnostic\JsDirectoryWritePermissionCheck'),
+    ),
+    'diagnostics.informational' => array(
+        // adds an informational system check
+        DI\get('Piwik\Plugins\MyPluginName\Diagnostic\JsDirectoryWritePermissionCheck'),
+    )
+];
```
sgiehl: Even though it seems to work without the `DI\add`, according to the PHP DI docs, the array would be overwritten  without it. So guess we should add it...

```suggestion
return [
    'diagnostics.required' => DI\add(array(
        // adds a required system check
        DI\get('Piwik\Plugins\MyPluginName\Diagnostic\JsDirectoryWritePermissionCheck'),
    )),
    'diagnostics.optional' => DI\add(array(
        // adds an optional system check
        DI\get('Piwik\Plugins\MyPluginName\Diagnostic\JsDirectoryWritePermissionCheck'),
    )),
    'diagnostics.informational' => DI\add(array(
        // adds an informational system check
        DI\get('Piwik\Plugins\MyPluginName\Diagnostic\JsDirectoryWritePermissionCheck'),
    ))
];
```

docs/4.x/system-check.md
```diff
@@ -0,0 +1,137 @@
...
+
+## Removing a system check
+
+You can remove existing system checks from other plugins using the `config/config.php` within your plugin where you define any system check class that should be removed like this:
+
+```php
+return [
+    'diagnostics.disabled' => array(
+        DI\get('Piwik\Plugins\Diagnostics\Diagnostic\PhpVersionCheck'),
+    ),
```
sgiehl:
```suggestion
    'diagnostics.disabled' => DI\add(array(
        DI\get('Piwik\Plugins\Diagnostics\Diagnostic\PhpVersionCheck'),
    )),
```

sgiehl (changes requested):

docs/4.x/system-check.md
```diff
@@ -0,0 +1,137 @@
+---
+category: Develop
+title: System check
+---
+# Enriching the system check
+
+Some plugins may only work when certain extra conditions are met, like having an unusual PHP extension installed or needing certain write permission on the file system. In these cases it is recommended to add new checks to the Matomo diagnostics system check which you can find by going to `Matomo Administration -> Diagnostic -> System Check`. 
+
+## Types of system checks
```
diosmosis, line 9: Given the code uses Diagnostic instead of "check", have you thought about calling them "diagnostics" here? As in `Types of diagnostics`

docs/4.x/system-check.md
```diff
@@ -0,0 +1,137 @@
...
+
+### An informational check.
+
+Adding an informational output is typically useful for the plugin developer themselves to troubleshoot issues. Such information can then be easily requested from the user when a bug report is made. The user can then provide this information by simply copy / pasting the system check.
+
+Typical information could be version numbers of installed packages or information about stored data.
+
+## Creating a new system check
+
+To create a new system check, use the [console](/guides/piwik-on-the-command-line) to execute below command (requires Matomo 4.4.1):
```
diosmosis:
```suggestion
To create a new system check, use the [console](/guides/piwik-on-the-command-line) to execute the below command (requires Matomo 4.4.1):
```

docs/4.x/system-check.md
diosmosis: Actually if the command is named `generate:system-check` then it doesn't matter, ignore this

docs/4.x/system-check.md
```diff
@@ -0,0 +1,137 @@
...
+
+```bash
+./console generate:system-check
+```
+
+The command will ask for your plugin name and for the name of the system check you want to create. You should enter some meaningful name that describes the check well. For example, "JS Directory Write Permission Check".
+
+Once all information is provided, a system check class will be created in the `Diagnostic` directory of your plugin containing an example on how to define a system check. In above example it would have created a file `Diagnostic/JsDirectoryWritePermissionCheck.php` within the chosen plugin.
```
diosmosis:
```suggestion
Once all information is provided, a system check class will be created in the `Diagnostic` directory of your plugin containing an example on how to define a system check. In the above example it would have created a file `Diagnostic/JsDirectoryWritePermissionCheck.php` within the chosen plugin.
```

docs/4.x/system-check.md
```diff
@@ -0,0 +1,137 @@
...
+    return $result;
+}
+```
+
+#### Statuses of system check
+
+Each check can have one of four different statuses:
+
+* **OK** (`DiagnosticResult::STATUS_OK`) - Everything is fine and the system passes the check.
+* **Error** (`DiagnosticResult::STATUS_ERROR`) - There is an error which does need to be fixed. Typically used for a required check.
```
diosmosis:
```suggestion
* **Error** (`DiagnosticResult::STATUS_ERROR`) - There is an error which needs to be fixed. Typically used for a required check.
```

diosmosis:
left a couple small suggestions, other than @sgiehl's items, everything looks good!

docs/4.x/system-check.md
tsteur: good one 👍

PR #500, opened 2021-08-04
Some basic documentation around data tables

tsteur:
Eventually will need to create a more "In depth" documentation for core developers. Figured it's good to have as a basic kind of documentation as I found it bit confusing in the beginning and at least it gives one a bit of an idea what they are about.

### DIFF
--- a/docs/4.x/datatable.md
+++ b/docs/4.x/datatable.md
@@ -26,7 +26,7 @@ foreach ($table->getRows() as $row) {
         echo $column . ': ' . $value; // would print eg label:1 nb_visits:1
     }
 }
-// Tables also implement the ArrayAccess interface and thereofore this works too:
+// Tables also implement the ArrayAccess interface and therefore this works too:
 foreach ($table as $row) {
     foreach ($row as $column => $value) {
         echo $column . ': ' . $value; 
--- a/docs/4.x/datatable.md
+++ b/docs/4.x/datatable.md
@@ -46,7 +46,7 @@ Each table may store multiple [rows](https://developer.matomo.org/api-reference/
 
 ## Manipulating data tables
 
-You can manipulate data tables in any kind of way by using filters:
+You can manipulate data tables in any kind of way by using *filters*, which are just callables that operate on a `DataTable` instance:
 
 * `$table->filter()` - Executes the filter right away. Depending on what the filter does, it may or may not be also applied to subtables. 
 * `$table->filterSubtables()` - Executes the filter right away. The filter won't be executed on the rows within this data table but only on subtables.
--- a/docs/4.x/datatable.md
+++ b/docs/4.x/datatable.md
@@ -40,7 +40,7 @@ There are three different types of data tables:
 
 * [DataTable](https://developer.matomo.org/api-reference/Piwik/DataTable) - Regular data table for reports that have a dimension. In that case, each row has typically a "label" column for the row's associated dimension value. 
 * [DataTable\Simple](https://developer.matomo.org/api-reference/Piwik/DataTable/Simple) - Used for reports that don't have a dimension. In that case, there is no "label" column and it only stores data for metrics. Such a table typically has only one row.
-* [DataTable\Map](https://developer.matomo.org/api-reference/Piwik/DataTable/Map) - Stores a set of regular or simple data tables when multiple periods are requested (think of `&period=day&date=last7`). In that case you can iterate over each table within the map using the `getDataTables()` method. The array key for each table stores the date which the subtable represents.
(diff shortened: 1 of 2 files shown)


diosmosis:
FYI, some of this might be duplicated in the class doc for `DataTable`.

docs/4.x/datatable.md
```diff
@@ -0,0 +1,153 @@
...
+category: DevelopInDepth
+title: Data Table
+---
+# Data Tables
+
+Data tables are used to store reporting data. [Visualisation](/guides/visualizing-report-data) decide how these tables are displayed in the UI and they can be requested in different formats using the [HTTP Reporting API](/api-reference/reporting-api) (json, xml, csv, ...).
+
+Data tables look just like a regular table:
+
+* A table has no or multiple rows
```
diosmosis:
```suggestion
* A table has a set of rows (which can be empty)
```

docs/4.x/datatable.md
```diff
@@ -0,0 +1,153 @@
...
+* Each row has multiple columns
```
diosmosis:
```suggestion
* Each row has multiple columns where the name and value of the column are both stored in the row
```

docs/4.x/datatable.md
```diff
@@ -0,0 +1,153 @@
...
+$table->addRowsFromSimpleArray(array(
+    array('label' => 'thing1', 'nb_visits' => 1, 'nb_actions' => 1),
+    array('label' => 'thing2', 'nb_visits' => 2, 'nb_actions' => 2)
+));
+foreach ($table->getRows() as $row) {
+    foreach ($row->getColumns() as $column => $value) {
+        echo $column . ': ' . $value; // would print eg label:1 nb_visits:1
+    }
+}
+// Tables also implement the ArrayAccess interface and thereofore this works too:
```
diosmosis:
```suggestion
// Tables also implement the ArrayAccess interface and therefore this works too:
```

docs/4.x/datatable.md
```diff
@@ -0,0 +1,153 @@
...
+        echo $column . ': ' . $value; 
+    }
+}
+```
+
+## Different types of data tables
+
+There are three different types of data tables:
+
+* [DataTable](https://developer.matomo.org/api-reference/Piwik/DataTable) - Regular data table for reports that have a dimension. In that case, each row has typically a "label" column.
```
diosmosis:
```suggestion
* [DataTable](https://developer.matomo.org/api-reference/Piwik/DataTable) - Regular data table for reports that have a dimension. In that case, each row has typically a "label" column for the row's associated dimension value. 
```

docs/4.x/datatable.md
```diff
@@ -0,0 +1,153 @@
...
+}
+```
+
+## Different types of data tables
+
+There are three different types of data tables:
+
+* [DataTable](https://developer.matomo.org/api-reference/Piwik/DataTable) - Regular data table for reports that have a dimension. In that case, each row has typically a "label" column. 
+* [DataTable\Simple](https://developer.matomo.org/api-reference/Piwik/DataTable/Simple) - Used for reports that don't have a dimension. In that case, there is no "label" column and it only stores data for metrics. Such a table typically has only one row.
+* [DataTable\Map](https://developer.matomo.org/api-reference/Piwik/DataTable/Map) - Stores a set of regular or simple data tables when multiple periods are requested (think of `&period=day&date=last7`). In that case you can iterate over each table within the map using the `getDataTables()` method. The array key for each table stores the date which the subtable represents.
```
diosmosis: Maps can also contain other maps, this could be worth mentioning?

docs/4.x/datatable.md
```diff
@@ -0,0 +1,153 @@
...
+
+Each table may store multiple [rows](https://developer.matomo.org/api-reference/Piwik/DataTable/Row).
+
+## Manipulating data tables
+
+You can manipulate data tables in any kind of way by using filters:
```
diosmosis:
```suggestion
You can manipulate data tables in any kind of way by using *filters*, which are just callables that operate on a `DataTable` instance:
```

docs/4.x/datatable.md
```diff
@@ -0,0 +1,153 @@
...
+
+* `$table->filter()` - Executes the filter right away. Depending on what the filter does, it may or may not be also applied to subtables. 
+* `$table->filterSubtables()` - Executes the filter right away. The filter won't be executed on the rows within this data table but only on subtables.
+* `$table->queueFilter()` - Similar to `filter()` but will be executed only after all the not needed rows have been removed.
+* `$table->queueFilterSubtables()` - Similar to `filterSubtables()` but will be executed only after all the not needed rows have been removed.
```
diosmosis, line 54: Would it be worth mentioning that queued filters are only applied when `applyQueuedFilters()` is called? And that this is called when processing a report for the API or for a visualization?

diosmosis (approved):
Left a couple suggestions otherwise looks good.

docs/4.x/datatable.md
tsteur: Didn't even know that (or didn't remember that). Will add it 👍

docs/4.x/datatable.md
tsteur: 👍  added some information around this

PR #508, opened 2021-08-10
Add link to Matomo setup

mwithheld:
Reduce setup issues for devlopers new to Matomo.

### Description:

Reduce setup issues for devlopers new to Matomo.


tsteur:
👍  Thanks for this @mwithheld

PR #516, opened 2021-08-16
Update piwiks-ini-configuration.md

geekdenz:
Add Boolean Configuration test documentation.

### Description:

Added documentation as suggest in

https://github.com/matomo-org/matomo/pull/17865#pullrequestreview-727190575 

### Review

* [ ] Functional review done
* [ ] Potential edge cases thought about (behavior of the code with strange input, with strange internal state or possible interactions with other Matomo subsystems)
* [ ] Usability review done (is anything maybe unclear or think about anything that would cause people to reach out to support)
* [ ] Security review done [see checklist](https://developer.matomo.org/guides/security-in-piwik#checklist)
* [ ] Code review done
* [ ] Tests were added if useful/possible
* [ ] Reviewed for breaking changes
* [ ] Developer changelog updated if needed
* [ ] Documentation added if needed

### DIFF
--- a/docs/4.x/piwiks-ini-configuration.md
+++ b/docs/4.x/piwiks-ini-configuration.md
@@ -69,6 +69,6 @@ If you want to make your plugin configurable, create a [Plugin Setting](/guides/
 
 For example `force_ssl = 1` is a boolean value in the configuration. By convention we use either 0 or 1 exactly for the value. It was suggested that we [improve and clarify this in the future](https://github.com/matomo-org/matomo/issues/17876).
 
-For INI files we use [parse_ini_string](https://www.php.net/manual/en/function.parse-ini-string.php), which means settings like `setting = on` and `setting = yes` also work as expected. The value usually gets checked against `== 1`, but is also checked for being truthy in various places.
+Settings like `setting = on`, `setting = off` and `setting = yes` etc. also work as expected. The value usually gets checked against `== 1`, but is also checked for being truthy in various places.
 
 See [ConfigTest.php](https://github.com/matomo-org/matomo/blob/2a06b5a99a8ce9ffd84b57a88bcc0f1513f13190/tests/PHPUnit/Unit/Config/ConfigTest.php#L52-L63).
--- a/docs/4.x/piwiks-ini-configuration.md
+++ b/docs/4.x/piwiks-ini-configuration.md
@@ -67,8 +67,28 @@ If you want to make your plugin configurable, create a [Plugin Setting](/guides/
 
 ## Boolean Configuration Options
 
-For example `force_ssl = 1` is a boolean value in the configuration. By convention we use either 0 or 1 exactly for the value. It was suggested that we [improve and clarify this in the future](https://github.com/matomo-org/matomo/issues/17876).
+For example `force_ssl = 1` is a boolean value in the configuration.
+By convention we check whether a feature is enabled by comparing
+the setting against the value `1` like this:
 
-Settings like `setting = on`, `setting = off` and `setting = yes` etc. also work as expected. The value usually gets checked against `== 1`, but is also checked for being truthy in various places.
+```php
+if ( \Piwik\Config::getInstance()->General['setting'] == 1) {
+}
+```
 
-See [ConfigTest.php](https://github.com/matomo-org/matomo/blob/2a06b5a99a8ce9ffd84b57a88bcc0f1513f13190/tests/PHPUnit/Unit/Config/ConfigTest.php#L52-L63).
+This is currently not consistent throughout the code base as some
+places compared the value as a boolean like below (not recommended):
+
+```php
+if ( \Piwik\Config::getInstance()->General['setting']) {
+}
+```
+
+It was suggested that we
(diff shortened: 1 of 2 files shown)


docs/4.x/piwiks-ini-configuration.md
```diff
@@ -64,3 +64,11 @@ Config::getInstance()->forceSave();
 **Plugins cannot add new configuration options.** If you are creating a core contribution and want to add a new INI option, you can simply add the option and its default value to `global.ini.php`.
 
 If you want to make your plugin configurable, create a [Plugin Setting](/guides/plugin-settings).
+
+## Boolean Configuration Options
+
+For example `force_ssl = 1` is a boolean value in the configuration. By convention we use either 0 or 1 exactly for the value. It was suggested that we improve and clarify this in the future.
```
tsteur: could add a link here? `[improve and clarify this in the future](https://github.com/matomo-org/matomo/issues/17876).` --> That's the issue I created for it.

docs/4.x/piwiks-ini-configuration.md
```diff
@@ -64,3 +64,11 @@ Config::getInstance()->forceSave();
...
+
+For INI files we use [parse_ini_string](https://www.php.net/manual/en/function.parse-ini-string.php), which means settings like `setting = on` and `setting = yes` also work as expected. The value gets converted to `=== 1`.
```
tsteur: btw not sure if we usually check with `== 1`?

tsteur (commented):
nice to see this documented 👍

tsteur:
@geekdenz could you have a look at the comments so we can merge it?

docs/4.x/piwiks-ini-configuration.md
```diff
@@ -64,3 +64,11 @@ Config::getInstance()->forceSave();
 **Plugins cannot add new configuration options.** If you are creating a core contribution and want to add a new INI option, you can simply add the option and its default value to `global.ini.php`.
 
 If you want to make your plugin configurable, create a [Plugin Setting](/guides/plugin-settings).
+
+## Boolean Configuration Options
+
+For example `force_ssl = 1` is a boolean value in the configuration. By convention we use either 0 or 1 exactly for the value. It was suggested that we [improve and clarify this in the future](https://github.com/matomo-org/matomo/issues/17876).
+
+For INI files we use [parse_ini_string](https://www.php.net/manual/en/function.parse-ini-string.php), which means settings like `setting = on` and `setting = yes` also work as expected. The value usually gets checked against `== 1`, but is also checked for being truthy in various places.
```
Findus23: > For INI files we use parse_ini_string

Just FYI: This isn't completely true. If the function is not available a PHP-implementation is used that might sometimes work a bit differently.
https://github.com/matomo-org/component-ini/blob/14c8cf5e78c8046a3e1b3f2fe8dc6cb474bc3f6b/src/IniReader.php#L75-L88

geekdenz: Thanks @Findus23 !

Do you have a suggestion how to alter the text or should we just delete it?

tsteur: @geekdenz maybe just remove the information that we use `parse_ini_string` and only mention that on and yes etc will be converted to 1 to.

geekdenz: I removed the reference to parse_ini_string and added some words.

docs/4.x/piwiks-ini-configuration.md
```diff
@@ -64,3 +64,11 @@ Config::getInstance()->forceSave();
 **Plugins cannot add new configuration options.** If you are creating a core contribution and want to add a new INI option, you can simply add the option and its default value to `global.ini.php`.
 
 If you want to make your plugin configurable, create a [Plugin Setting](/guides/plugin-settings).
+
+## Boolean Configuration Options
+
+For example `force_ssl = 1` is a boolean value in the configuration. By convention we use either 0 or 1 exactly for the value. It was suggested that we [improve and clarify this in the future](https://github.com/matomo-org/matomo/issues/17876).
+
+Settings like `setting = on`, `setting = off` and `setting = yes` etc. also work as expected. The value usually gets checked against `== 1`, but is also checked for being truthy in various places.
```
tsteur: @geekdenz the second paragraph is not quite clear. Maybe something like below would be a bit more clear?

For example `force_ssl = 1` is a boolean value in the configuration. By convention we check whether a feature is enabled by comparing the setting against the value `1` like this:

```php
if ( \Piwik\Config::getInstance()->General['setting'] == 1) {
}
```

This is currently not consistent throughout the code base as some places compared the value as a boolean like below (not recommended):

```php
if ( \Piwik\Config::getInstance()->General['setting']) {
}
```

It was suggested that we [improve and clarify this in the future](https://github.com/matomo-org/matomo/issues/17876). We can't refactor the boolean checks to compare against `1` as it would be a breaking change since it would mean if someone configures for example `2` then it would no longer be interpreted as enabled.

Please note when someone configures `setting = on` or `setting = yes` then their value gets converted in the code to `1` as well. The values `off` and `no` will be converted to `0`.

PR #534, opened 2021-08-25
Change order in InDepth menu to show understanding Matomo first

tsteur:
Currently the menu looks like this:
image

I thought it might make sense to mention the pages "Understanding Matomo" first before "Core development" as it may be more important and more relevant to more devs see screenshot below.

Just a random thought. Happy to close without merging.

image


sgiehl (approved):
LGTM

PR #535, opened 2021-08-26
Updates for the git guide

tsteur:
### Review

### DIFF
--- a/docs/git.md
+++ b/docs/git.md
@@ -3,7 +3,9 @@ category: DevelopInDepth
 ---
 # Git
 
-At Matomo we're using Git version control system. Below are some tips for using git.
+At Matomo we're using Git version control system. Below are some tips for using git. Below is a guide for how we work using git on the command line.
+
+If you prefer using Git through your IDE like PHPStorm, or if you are more accustomed to use a slightly different flow then please go with your preferred option of what you feel comfortable with. As often there is no right or wrong and not one way to do something.
 
 ## Prerequisites
 
@@ -108,25 +110,32 @@ Pushing means it will send all the previous commits to the remote server which i
 
 When you visit the repository you pushed the change to on github.com  then GitHub will directly show you branch you pushed to on the main page and you can click on it and inspect the changes. You then either make more pushes or if you are happy with the changes or seek feedback then you [create a pull request](/guides/pull-request-reviews).
 
+#### Force pushes
+
+Most main branches are protected from force pushes that rewrite the history. If you ever need to rewrite the history (which should be needed rarely) then you will want to use the below command:
+
+`git push --force-with-lease`
+
+While `git push --force` works too it overwrites the entire remote branch with your local branch, `--force-with-lease` is safer as it makes sure you won't overwrite someone else's work.
+
 ## Submodules
 
-### Accidentally Commited Submodules?
+### Accidentally Committed Submodules?
 
 Have you done some work on Matomo or another project and accidentally overridden the submodule commit?
 
 This may help you find your commands to recover.
 
 ```
-git checkout 4.x-dev             # main branch, sometimes also called master
+# below replace `$MAIN_BRANCH` with the name of the main branch. Sometimes this is "live", "develop" or `*.x-dev` eg `4.x-ev`.
+# and repace "$SUBMODULE_DIR" with a path to the submodule you want to correct. For example "plugins/Morpheus/icons".
+git checkout $MAIN_BRANCH             # switch to the main branch
 git submodule init
 git submodule update --recursive
-git checkout m-17640             # your branch where you did the wrong commit
-git checkout 4.x-dev plugins/Morpheus/icons
-                                 # main branch
-                                 # submodule that was wrong
-git add plugins/Morpheus/icons   # add the submodule back in its original state
-git commit -m 'revert git submodule to 4.x-dev version #17640'
-                                 # commit the changes
+git checkout $FEATURE_BRANCH             # replace `$FEATURE_BRANCH` with the name of your branch where you did the wrong commit
+git checkout $MAIN_BRANCH $SUBMODULE_DIR 
+git add $SUBMODULE_DIR   # add the submodule back in its original state
+git commit -m 'revert git submodule' # commit the changes
 ```
 
 ## Other useful commands
--- a/docs/git.md
+++ b/docs/git.md
@@ -5,41 +5,108 @@ category: DevelopInDepth
 
 At Matomo we're using Git version control system. Below are some tips for using git.
 
-## Commits
+## Prerequisites
 
-### Undo last commit that is not pushed yet without losing the changes
+* [Git LFS needs to be installed](https://github.com/git-lfs/git-lfs/wiki/Installation)
+
+## Cloning a repository
+
+Github supports cloning via HTTPS and SSH URLs. You will always want to use SSH URLs when cloning a repository, for example `git@github.com:matomo-org/matomo.git`
+
+To clone a repository use for example:
+
+`git clone git@github.com:matomo-org/matomo.git matomo`
+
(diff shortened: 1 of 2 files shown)


docs/git.md
```diff
@@ -3,61 +3,145 @@ category: DevelopInDepth
...
+
+When you visit the repository you pushed the change to on github.com  then GitHub will directly show you branch you pushed to on the main page and you can click on it and inspect the changes. You then either make more pushes or if you are happy with the changes or seek feedback then you [create a pull request](/guides/pull-request-reviews).
+
+#### Force pushes
+
+Most main branches are protected from force pushes that rewrite the history. If you ever need to rewrite the history (which should be needed rarely) then you will want to use the below command:
+
+`git push --force-with-lease`
+
+While `git push --force` works too it overwrites the entire remote branch with your local branch, `--force-with-lease` is safer as it makes sure you won't overwrite someone else's work.
```
sgiehl: That's actually not fully correct. `force-with-lease` actually only check if you local copy of origin is the same as the remote origin. So it actually prevents pushing if someone else pushed the branch in the meantime. But a simple `git fetch; git push --force-with-lease` would then work and "overwrite" the remote changes.

docs/git.md
```diff
@@ -3,61 +3,145 @@ category: DevelopInDepth
...
+You don't have to add all files at once. You can add them over multiple comments before committing them:
+
+```bash
+git add index.php 
+git add lang-directory file2
+```
+
+In some cases you might prefer the following command which asks you interactively which changes you want to add or not:
+
+`git add -p`
```
sgiehl: Note: This won't work for binary files. So for example images would be skipped by this command

docs/git.md
```diff
@@ -3,61 +3,145 @@ category: DevelopInDepth
 ---
 # Git
 
-At Matomo we're using Git version control system. Below are some tips for using git.
+At Matomo we're using Git version control system. Below are some tips for using git. Below is a guide for how we work using git on the command line.
```
sgiehl: Sounds a bit weird with `Below are... Below is...` Maybe something like this sounds better?
```suggestion
At Matomo we're using Git version control system. Below are some tips for using git along with a guide for how we work using git on the command line.
```

docs/git.md
```diff
@@ -3,61 +3,145 @@ category: DevelopInDepth
...
 
-## Commits
+If you prefer using Git through your IDE like PHPStorm, or if you are more accustomed to use a slightly different flow then please go with your preferred option of what you feel comfortable with. As often there is no right or wrong and not one way to do something.
 
-### Undo last commit that is not pushed yet without losing the changes
+## Prerequisites
 
-```git reset --soft HEAD~1```
+* [Git LFS needs to be installed](https://github.com/git-lfs/git-lfs/wiki/Installation)
```
sgiehl, line 12: Should we mention that by default the lfs files won't be checked out, and that this is only needed for writing or updating UI tests. See https://developer.matomo.org/guides/tests-ui#requirements

docs/git.md
```diff
@@ -3,61 +3,145 @@ category: DevelopInDepth
...
+## Cloning a repository
 
-### Patch Add
+Github supports cloning via HTTPS and SSH URLs. You will always want to use SSH URLs when cloning a repository, for example `git@github.com:matomo-org/matomo.git`
 
-```git add -p```
+To clone a repository use for example:
 
-With this you can add your changes interactively.
+`git clone git@github.com:matomo-org/matomo.git matomo`
```
sgiehl: Would it make sense to mention the submodules here and that a `git submodule init` is needed to initialize and check them out?

docs/git.md
```diff
@@ -3,61 +3,145 @@ category: DevelopInDepth
...
-### Log Changes
+This would clone the Matomo repository into a `matomo` directory.
 
-```git log -p```
+## Creating a new branch
 
-You can log all changes that have happened to see exactly what was changed when and by who.
+Before you start coding on a new feature, you should make sure to keep your changes separate from the main branch. This will allow you to [create a pull request from this branch later which can then be reviewed](/guides/pull-request-reviews).
+
+We'll give our new branch a name. The branch name ideally always contains the GitHub issue number eg `1111` or `m1111` or if it's a Jira issue then eg `dev-1111`. Optionally, a descriptive name can be added if wanted. It's not a requirement though since the branch usually exists only temporarily anyway and a PR with description etc often exists too. To add a new branch, run the following command:
```
sgiehl: not sure if we maybe should avoid using the issue number as branch name only. At some point this might result in some weird stuff when an issue number matches a short version of a git hash. Then git checkout might be ambiguous

docs/git.md
```diff
@@ -3,61 +3,145 @@ category: DevelopInDepth
...
+git checkout -b bugfix $BASE_BRANCH
+```
+
+Where you need to replace `$BASE_BRANCH` with the current main branch. This is for example `*.x-dev` when working on Matomo On-Premise (`4.x-dev` is the main branch at the time of writing this). For Matomo for WordPress this would be `develop`. For example: `git checkout -b bugfix 4.x-dev`. This way it will create a new branch from the `4.x-dev` branch. Otherwise, if you don't specify the base branch and you are already in another branch that has changes, then it would copy all changes from the current branch into the new branch.
+
+The checkout command will create a new branch if the `-b` option is supplied, otherwise it will try to load an existing branch.
+
+## Adding, committing and pushing code
+
+After you added or modified a few files you will want to push these changes to the remote repository. It's good to commit regularly even if a feature is not finished yet so you always have a backup if changes get lost for some reason.
```
sgiehl:
```suggestion
After you added or modified a few files you will want to push these changes to the remote repository. It's good to commit and push regularly even if a feature is not finished yet so you always have a backup if changes get lost for some reason.
```

docs/git.md
```diff
@@ -3,61 +3,145 @@ category: DevelopInDepth
...
+git commit -m 'My awesome commit message'
 ```
-git bisect start
-git bisect good
-git bisect bad
+
+If you executed the commit and notice a typo then you can update the commit message using the `--amend` option:
+
+```bash
+git commit --amend -m 'My awesome commit message'
```
sgiehl, line 105: Should we mention that a force push is needed when amending a commit that was already pushed?

docs/git.md
```diff
@@ -3,61 +3,145 @@ category: DevelopInDepth
...
-This is the most efficient way to find a commit that introduced a bug.
+In Matomo the commit messages are less important as we are usually creating a pull request and then squash all the commits into one commit when merging the PR. The title of the pull request will become the commit message.
+
+When you make a commit then these commits are not yet sent to any remote server. You can do one or multiple commits before actually pushing these files to a remote server which we will go into the next step.
+
+### Reviewing the last commit
+
+If you committed changes and want to see what the last commit looked like then you can execute `git show`.
+
+#### Undo last commit that is not pushed yet without losing the changes
```
sgiehl, line 129: This can also be used to undo a commit that had been pushed, but it will then require a force push

sgiehl (changes requested):

docs/git.md
tsteur: they used to be cloned by default or not? Are you saying after `git clone` we now need to execute `git lfs pull --exclude=` everywhere including the submodules?

docs/git.md
tsteur: I'll add a note to not amend commits after pushing

tsteur:
Cheers @sgiehl I'm still in the process of writing the guide but incorporated some things already

sgiehl:
For most submodules the UI files aren't stored on LFS. But yes this needs
to be executed at least in matomo repo as our gitconfig prevent the
checkout by default.

justinvelluppillai (approved):
Looks good, a lot of nice tips in here. I wonder if we could produce a one page cheat sheet along the same lines at some point.

PR #545, opened 2021-09-06
document "How many rows should I expect in an archive table?"

tsteur:


docs/4.x/database-schema.md
```diff
@@ -307,6 +307,16 @@ In `archive_blob` tables:
 - the `index_period_archived` index is used in the same way as the one in `archive_numeric` tables
 - `archive_blob` tables do not have an index that makes it fast to query for rows by site, period and archived date. This is because they should not be queried this way. Instead, the `archive_numeric` table should be queried and the `idarchive` values saved. These values can be used to query data in the `archive_blob` table.
 
+### How many rows should I expect in an archive table?
+
+Seeing millions of rows in an archiving table is not uncommon. Depending on what features a Matomo has enabled and how many unique goals, custom reports, funnels, action URLs and referrers a site tracks each individidual archive contains on average approximately 30-100 entries in a blob table and 10-30 entries on average for each archive in the numeric table. Some Matomos might store a lot more or a lot less rows per archive.
```
sgiehl:
```suggestion
Seeing millions of rows in an archiving table is not uncommon. Depending on what features a Matomo has enabled and how many unique goals, custom reports, funnels, action URLs and referrers a site tracks each individual archive contains on average approximately 30-100 entries in a blob table and 10-30 entries on average for each archive in the numeric table. Some Matomos might store a lot more or a lot less rows per archive.
```

sgiehl (changes requested):

PR #547, opened 2021-09-06
Documentation on composer requirements

sgiehl:
### Description:

We currently actually didn't really have some process for updating our composer dependencies.

I've created a new GitHub action in https://github.com/matomo-org/matomo/pull/17969 to automate minor & patch release updates.

For major updates we actually need to find a proper solution. Maybe another action could create issues if major updates are available, so we see that updates are available. Another option would be some kind of recurring task every X months to check for updates.

Might be good to discuss that, so we can adjust this documentation accordingly.

Let me know if anything else should be changed or is maybe missing.

### Review

### DIFF
--- a/app/helpers/Content/Category/DevelopInDepthCategory.php
+++ b/app/helpers/Content/Category/DevelopInDepthCategory.php
@@ -38,6 +38,7 @@ class DevelopInDepthCategory extends Category
                 new Guide('reproducing-issues'),
                 new Guide('core-faqs'),
                 new Guide('core-components'),
+                new Guide('composer-dependencies'),
                 new Guide('release-management'),
                 new RemoteLink('Matomo\'s Roadmap', 'https://matomo.org/roadmap/'),
             ]),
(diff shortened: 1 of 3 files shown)


tsteur:
@sgiehl can you add this guide to a menu item see https://github.com/matomo-org/developer-documentation#adding-the-new-guide-to-a-menu-item and then merge?

sgiehl:
@tsteur I've added it to the menu. Not sure if we should already merge it. It mentions and links an action which will be available after merging https://github.com/matomo-org/matomo/pull/17969, so merging this PR now, might introduce broken links

tsteur:
Looks like it's only one link? We could still merge or just remove that link and then merge.

geekdenz:
Seems good to merge from my pov.

sgiehl:
Removing the link only doesn't make much sense. The docs are describing a process that is available after https://github.com/matomo-org/matomo/pull/17969 was merged. Reviewing that one should be quite simple, as the action isn't doing much. Afterwards we can directly merge this one.

PR #548, opened 2021-09-08
Add basic DB guide

tsteur:
### Description:

Just wanted to document somewhere that a query with too many place holders can be very slow. Noticed then there wasn't really a good place yet and we didn't even have a DB guide which is needed for most plugins. So spent 20 minutes and created this basic guide which will help developers in the right directions.

### Review

### DIFF
--- a/docs/4.x/database.md
+++ b/docs/4.x/database.md
@@ -45,4 +45,5 @@ $db->query($sql, array($key, $value, (int) $ttlInSeconds));
 
 * Always only fetch the data you need and make use of indexes. [See also our profiling the DB guide.](/guides/profiling-code#database-queries)
 * For security to prevent SQL injections, bind parameters. Instead of `WHERE userername = "$var"` use `WHERE userername = ?` and bind the value.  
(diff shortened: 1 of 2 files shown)


docs/4.x/database.md
```diff
@@ -0,0 +1,48 @@
...
+```php
+$sql = sprintf('INSERT INTO table_name (`key`, `value`, `expiry_time`) VALUES (?,?,(UNIX_TIMESTAMP() + ?))';
+$db->query($sql, array($key, $value, (int) $ttlInSeconds));
+```
+
+## Performance considerations
+
+* Always only fetch the data you need and make use of indexes. [See also our profiling the DB guide.](/guides/profiling-code#database-queries)
+* For security to prevent SQL injections, bind parameters. Instead of `WHERE userername = "$var"` use `WHERE userername = ?` and bind the value.  
+* If your query has hundreds or thousands of bound parameters, then this can make your query extremely slow. In that case, if possible, you will not want to use bound parameters. It is typically only possible if you can cast values to integers, as this way you can ensure there won't be a SQL injection. Example: `WHERE idvisit = (int)$idvisit`
```
sgiehl: Just a side note on that one: We even had issues in the past, where a query ended up with too many placeholders. There's a limit of 65,536 on placeholders. Especially when performing queries that handle a unknown amount of visits this can be easily reached.

tsteur: great, thanks. Will add it

PR #549, opened 2021-09-12
document setPagePerformanceTiming() function on core tracker

geekdenz:
### Description:

Relates to https://github.com/matomo-org/matomo/pull/17901 & https://github.com/matomo-org/matomo/issues/16859 . Please only merge once that is reviewed and merged.

### Review


docs/4.x/tracking-javascript.md
```diff
@@ -156,6 +156,7 @@ Piwik uses first party cookies to keep track of some user information over time.
...
 *   `setRequestQueueInterval( interval )` -  Defines after how many ms a queued requests will be executed after the request was queued initially. The higher the value the more tracking requests can be send together at once. `interval` has to be at least `1000` (1000ms = 1s) and defaults to 2.5 seconds.
+* `setPagePerformanceTiming(networkTimeInMs, serverTimeInMs, transferTimeInMs, domProcessingTimeInMs, domCompletionTimeInMs, onloadTimeInMs)` - Manually set performance metrics in ms (thousandth of a second like `setRequestQueueInternal`) in a SPA or when Matomo cannot detect some metrics. You can set parameters to `undefined` if you do not want to track this metric. Note that `undefined` is a valid value and will not throw an error. This is by design, because users may not want to set all parameters. At least one parameter needs to be set. It will only be used on the next page view and only tracked once.
```
tsteur: @geekdenz nice. 

You can remove `thousandth of a second like setRequestQueueInternal` as it's not needed. Maybe rather spell out `ms` -> milliseconds then it'll be clear.

It be also good to spell out SPA as some won't know what it means.
 
You could maybe also remove `Note that undefined is a valid value and will not throw an error. This is by design, because users may not want to set all parameters. ` as it's not really needed but also can keep it.

Btw we mark optional parameters with `[optionalParameterName]`. Meaning eg `[serverTimeInMs], [transferTimeInMs], `. It's a bit tricky here because at least one of these values is required but could simply mark them all as optional as you mention at least one needs to be defined.

`It will only be used on the next page view` might be better to clarify "it" like `The set performance timings will be tracked only on the next page view. If you track another page view then you may need to set the performance timings again.` something like it.

PR #560, opened 2021-09-30
Minor tweak for omnifixture

tsteur:
it should not be updated after every fixture change

### Review

### DIFF
--- a/docs/4.x/tests-system.md
+++ b/docs/4.x/tests-system.md
@@ -25,17 +25,15 @@ Fixtures are stored in [tests/PHPUnit/Fixtures](https://github.com/matomo-org/ma
 
 ### OmniFixture
 
-We also have an OmniFixture that includes all other Fixtures. OmniFixture is used for screenshot tests to provide data across most reports.
+We also have an OmniFixture that includes all other Fixtures. OmniFixture is used for screenshot tests to provide data across most reports. This fixture is created so the fixture has a lot of different data in the UI tests and it's reused and created as a SQL for speed. This is not neccessarily the best solution of how this is done and maybe in the future we rather want a dedicated UI test fixture or a dedicated UI test fixture for every test group instead of one big one.
 
 ### Keep OmniFixture up to date
 
-Remember to update the [Omnifixture SQL dump](https://github.com/matomo-org/matomo/blob/master/tests/resources/OmniFixture-dump.sql) whenever you make any change to any fixture. You can use:
+Rarely, we may need to update the [Omnifixture SQL dump](https://github.com/matomo-org/matomo/blob/master/tests/resources/OmniFixture-dump.sql) so a specific feature can be tested. You can use:
 
     ./console tests:setup-fixture OmniFixture --sqldump=OmniFixture-dump.sql
 
-Keeping the OmniFixture up to date makes it easier to see which tests fail after each small fixture change.
-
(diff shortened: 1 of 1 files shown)


sgiehl:
Actually it would be good to update it after every fixture change and even after every dimension change. Guess it would be helpful to have some kind of automatic process to create a PR with updated OmniFixture so you only need to check the changed images, which shouldn't be that many if it's done more often.
Currently creating a PR to update the OmniFixture would for sure consume a lot more time as it wasn't updated for a long time.

justinvelluppillai:
@tsteur @sgiehl is this to be merged or modified? The last comment seems to contain a query that hasn't been resolved yet.

tsteur:
I'm afraid that if we're updating it regularly after every change, there will often be many screenshot changes failing that will need to be updated. Especially the more devs there are and the more changes there are in the fixtures etc which could result in a lot of overhead and interruptions rather than just doing it once or twice a year kind of thing and then spending bit more time if needed. I don't have a big preference but can see it's likely causing overhead and more interruptions than it's worth it maybe.
Maybe in that case we would even need to think about a completely different solution so that omnifixture doesn't depend on other fixtures or only selected fixtures or something.

sgiehl:
In the last half year there were not more than three relevant changes to any fixture I can remember of. But I would agree that we should change the process or adjust the tests if updating the omnifixture would get too time consuming.

tsteur:
In this case I'm also thinking forward that we'll have more and more developers and this will be happening even more often in the future. It could end up that people try avoid changing fixtures so we don't need to spend time fixing these tests which is tedious. And even updating 3 times in last 6 months can cause a lot of work when done properly.

So maybe we would rather need to look at a different way of managing the omnifixture?

sgiehl:
Personally I think, if the omnifixture would be updated with each PR (where some fixture was changed/added), this would actually not be much more effort. When changing/adding a fixture in most cases some UI tests are changing nevertheless. So you already need to spend some time on updating the screenshots. Having an automated process that would update the omnifixture in a PR would be easy to implement. So it's only about updating a few more screenshots.

But I can also spend some time on thinking about other concepts how to automatically create a fixture that contains data for all cases, as it's important to not miss anything for the UI tests. Having the Omnifixture outdated that way already has the risk to not see possible regressions in our tests for some cases.

tsteur:
> So it's only about updating a few more screenshots.

The problem is that many screenshots might change and the numbers look different etc and for every single screenshot you need to investigate if the change is actually expected and still accurate. Checking what the raw data is in the DB, double checking the new numbers are shown in the new screenshot are still correct. This can easily take many hours every time if we want to ensure that the tests are still accurate and that the shown results are expected. Meaning each screenshot will take quite a bit of time to investigate. Sometimes that will be easier depending on what fixture changes, sometimes it'll take a lot longer. Of course when there's only a simple change in a fixture then it can be faster to do it right away, if it impacts many screenshots or tracks more data somewhere then it can take much more time to update it every time. That's unless we say the data in the UI tests is irrelevant and it's ok if they are wrong as they are tested by system tests and we only look at the functionality that the screenshots tests such as opening a subtable, flattening, ...

tsteur:
@sgiehl any further thoughts? Or should we maybe turn this into a new issue to find a solution?

sgiehl:
Creating a new issue sounds good.

tsteur:
👍  created https://github.com/matomo-org/matomo/issues/19660 and will close this one.

PR #611, opened 2022-02-10
Adds new guide about how to handle request parameters

sgiehl:
### Description:

Let me know if there are any questions or feedback.

refs https://github.com/matomo-org/matomo/issues/18719#issuecomment-1034426019

### Review

### DIFF
--- a/docs/4.x/request-parameters.md
+++ b/docs/4.x/request-parameters.md
@@ -1,0 +1,73 @@
+---
+category: Develop
+title: Handling Request Parameters
+---
+# Handling Request Parameters
+
+If you need to access a query or post parameter value, use the [`Common::getRequestVar()`](/api-reference/Piwik/Common#getrequestvar) method.
+
+**To avoid XSS vulnerabilities, never access `$_GET`/`$_POST` directly, always go through [Common::getRequestVar](/api-reference/Piwik/Common#getrequestvar).**
+
+## Type handling
+
+When working with request variables you should be aware that your variables might not be provided with the type you expect.
+Someone could for example provide `&var[]=x` in the URL instead of `&var=x`, causing a variable to be of type `array` instead of `string`.
+
+When using `$_GET` or `$_POST` directly your variables can actually only have the two types: `string` or `array`.
+`Common::getRequestVar()` is able to validate your variables for some other types as well. Internally this is done by checking if casting the variable changes it's value.
+
+To force type validation you can pass one of the following types as third parameter:
+- `string`
+- `array`
+- `int` / `integer`
+- `float`
+- `json` (this tries to perform a `json_decode` on the provided value)
+
+It's recommended to use the method with three parameters where ever a certain type is expected. This will prevent PHP warnings or errors if an unexpected type was provided.
+
+## Distinguish between GET and POST
+
+In some cases it might be needed to ensure a parameter was provided by either GET or POST. You can achieve that by passing `$_GET` or `$_POST` as fourth parameter.
+By default, a combination of `$_GET` and `$_POST` will be used, while GET parameters will overwrite POST parameters.
+
+To receive a variable from POST only you can e.g. use:
+
+```php
+$var = Common::getRequestVar('var', null, 'string', $_POST);
+```
+
+## Usage Examples
+
+### Request parameter in a certain type required:
+
+```php
+$var = Common::getRequestVar('var', null, 'string');
+```
+
+This throws an exception if the variable is not provided or has another type than String
+
+### Request parameter can be provided in a certain type, but a default should be used if not:
+
+```php
+$var = Common::getRequestVar('var', $default, 'string');
+```
+
+This uses $default if the variable is not provided or has another type than String
+
+### Request parameter is required but might have any type:
+
+```php
+$var = Common::getRequestVar('var');
+```
+
+This throws an exception if the variable is not provided, otherwise the variable can have any type.
+If you use this please ensure to do the type handling within your code, so the code won't break if an Array might be provided instead of a String.
+
+### Request parameter can be provided, but a default should be used if not:
+
+```php
+$var = Common::getRequestVar('var', $default);
+```
+
+This won't throw an exception in any case. But you can't expect the variable to have a certain type.
+If you use this please ensure to do the type handling within your code, so the code won't break if an Array might be provided instead of a String.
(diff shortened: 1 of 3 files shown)


docs/4.x/request-parameters.md
```diff
@@ -0,0 +1,73 @@
...
+
+## Usage Examples
+
+### Request parameter in a certain type required:
+
+```php
+$var = Common::getRequestVar('var', null, 'string');
+```
+
+This throws an exception if the variable is not provided or has another type than String
```
tsteur: `if the variable is not provided` should be maybe `if the request parameter is not provided`

docs/4.x/request-parameters.md
```diff
@@ -0,0 +1,73 @@
+---
+category: Develop
+title: Handling Request Parameters
+---
+# Handling Request Parameters
+
+If you need to access a query or post parameter value, use the [`Common::getRequestVar()`](/api-reference/Piwik/Common#getrequestvar) method.
```
tsteur: Can we maybe show the full signature here with parameters? as in `getRequestVar($varName, $varDefault = null, $varType = null, $requestArrayToUse = null)`

tsteur (changes requested):
Can we mention the maybe unexpected that when default parameter is used, and a type is specified, that the default will be converted to that type? As in `default = false, 'string'` would return an empty string? Maybe also with an example? We might need to mention though that this again doesn't seem to be the case for `bool|float|json` :) 

We could also mention that the method changes the input and sanitizes it which can have various side effects and often they would want to do a `Common::unsanitizeInputValue` afterwards? Not sure if there are then still any side effects (likely).

There might be few other unexpected things. Like requesting a string but if it looks like a float we might change the input etc 
image

bx80 (approved):

peterhashair (approved):
LGTM

docs/4.x/request-parameters.md
sgiehl:
```suggestion
If you need to access a query or post parameter value, use the [`Common::getRequestVar($varName, $varDefault = null, $varType = null, $requestArrayToUse = null)`](/api-reference/Piwik/Common#getrequestvar) method.
```

docs/4.x/request-parameters.md
sgiehl:
```suggestion
This throws an exception if the request parameter is not provided or has another type than String
```

docs/4.x/request-parameters.md
```diff
@@ -0,0 +1,73 @@
...
+
+This throws an exception if the variable is not provided or has another type than String
+
+### Request parameter can be provided in a certain type, but a default should be used if not:
+
+```php
+$var = Common::getRequestVar('var', $default, 'string');
+```
+
+This uses $default if the variable is not provided or has another type than String
```
sgiehl:
```suggestion
This uses $default if the request parameter is not provided or has another type than String
```

docs/4.x/request-parameters.md
```diff
@@ -0,0 +1,73 @@
...
+
+### Request parameter is required but might have any type:
+
+```php
+$var = Common::getRequestVar('var');
+```
+
+This throws an exception if the variable is not provided, otherwise the variable can have any type.
```
sgiehl:
```suggestion
This throws an exception if the request parameter is not provided, otherwise the variable can have any type.
```

docs/4.x/request-parameters.md
```diff
@@ -0,0 +1,73 @@
...
+When using `$_GET` or `$_POST` directly your variables can actually only have the two types: `string` or `array`.
+`Common::getRequestVar()` is able to validate your variables for some other types as well. Internally this is done by checking if casting the variable changes it's value.
+
+To force type validation you can pass one of the following types as third parameter:
+- `string`
+- `array`
+- `int` / `integer`
+- `float`
+- `json` (this tries to perform a `json_decode` on the provided value)
+
```
sgiehl, line 27:
```suggestion

When providing a default value (other than `null`) together with a specific type, please ensure the default value has the same type. The default will be casted to the expected type, which might cause unexpected results. This doesn't apply for the special type `json`.

```

sgiehl:
> Can we mention the maybe unexpected that when default parameter is used, and a type is specified, that the default will be converted to that type? As in `default = false, 'string'` would return an empty string? Maybe also with an example? We might need to mention though that this again doesn't seem to be the case for `bool|float|json` :)

Add a suggestion to change that

> We could also mention that the method changes the input and sanitizes it which can have various side effects and often they would want to do a `Common::unsanitizeInputValue` afterwards? Not sure if there are then still any side effects (likely).

same for that one.

> There might be few other unexpected things. Like requesting a string but if it looks like a float we might change the input etc image

I don't think it's worth mentioning that at all. By default the method should be used with one of the super globals `$_GET` and/or `$_POST`. That results in all variables to be of type string or array. The additional checks above are so actually only relevant when passing a custom request array object, that contains variables with other types. As that should be a quite uncommon use case, I wouldn't mention it, so people don't get confused with that.

tsteur:
👍 
Not sure why but I can't see a suggestion for `unsanitizeInputValue`. Can you maybe double check (I might just not find it) and then commit the suggestion and merge it?

sgiehl:
Hm. I was pretty sure I added a suggestion for that. But seems it somehow got lost. Will add one again...

docs/4.x/request-parameters.md
```diff
@@ -0,0 +1,75 @@
+---
+category: Develop
+title: Handling Request Parameters
+---
+# Handling Request Parameters
+
+If you need to access a query or post parameter value, use the [`Common::getRequestVar($varName, $varDefault = null, $varType = null, $requestArrayToUse = null)`](/api-reference/Piwik/Common#getrequestvar) method.
+
+**To avoid XSS vulnerabilities, never access `$_GET`/`$_POST` directly, always go through [Common::getRequestVar](/api-reference/Piwik/Common#getrequestvar).**
+
```
sgiehl, line 10:
```suggestion

Using `Common::getRequestVar` will automatically sanitize the request variables using [Common#sanitizeInputValues](/api-reference/Piwik/Common#sanitizeinputvalues). For arrays and parsed json data this is done recursively. Internally this will escape e.g. special html characters. If you need to work with unescaped values, you may need to apply [Common#unsanitizeInputValues](/api-reference/Piwik/Common#unsanitizeinputvalues).

```

peterhashair: @sgiehl sorry I thought that was my PR, commit suggestion sorry about that .

tsteur (approved):

PR #617, opened 2022-03-02
Mention we need to update github action when we work on a new major release

tsteur:
Not sure when we would update that action but maybe when we start working on that new release.

There's a step `matomo-package build script` @sgiehl can this be removed now?
image

### Description:

Please include a description of this change and which issue it fixes. If no issue exists yet please include context and what problem it solves.

### Review

### DIFF
--- a/docs/release-management.md
+++ b/docs/release-management.md
@@ -83,10 +83,6 @@ See https://github.com/matomo-org/matomo-marketplace/wiki/New-Piwik-version
 
 ### When releasing a first beta
 
-#### matomo-package build script
-
-* [in `scripts/build.sh`](https://github.com/matomo-org/matomo-package/blob/master/scripts/build-package.sh), edit `CURRENT_LATEST_MAJOR_VERSION` and set it to the new major release eg. `4`
-
 #### developer.matomo.org
 
 * Follow steps as described in README.md: https://github.com/matomo-org/developer-documentation/#how-to-add-docs-for-a-new-matomo-version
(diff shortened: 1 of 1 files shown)


sgiehl:
> There's a step matomo-package build script @sgiehl can this be removed now?

@tsteur yes. that one can be fully removed. I have refactored all the code so it is no longer required to set the major version somewhere. The current major version is always detected by looking at https://builds.matomo.org/LATEST

Btw. the release action doesn't need to be adjusted neither. The check for valid branches matches again `[4-9].x-dev`. So as long as we don't want to release Matomo 10.0, no change will be needed there.

tsteur:
👍  removed that section, should be good to merge now or is there anything else we need to change @sgiehl ?

docs/release-management.md
```diff
@@ -63,6 +63,7 @@ Also, as soon as we start working on the next major version, we need to require
 * Travis CI, see what we did for Matomo 3 in https://github.com/matomo-org/matomo/pull/8452#discussion-diff-35731015L93 and https://github.com/matomo-org/travis-scripts/pull/53
 * Create the `PULL_REQUEST_TEMPLATE` similar to https://github.com/matomo-org/matomo/pull/12412/files
 * Once the build for 4.x-dev succeeds, make it the default branch for Matomo and all plugins
+* Update the [submodule github action](https://developer.matomo.org/guides/release-management#when-starting-to-work-on-the-next-major-release) to use the new main branch
```
sgiehl: Actually there are more than the `submodules update` action that need to be adjusted. But guess we can also have a look if it's easy to change the actions so they are using the configured default branch...

tsteur: @sgiehl which other ones would we need to adjust? We could simply document them here quick and then it be easy to adjust them when needed

sgiehl: [submodule action](https://github.com/matomo-org/matomo/blob/4.x-dev/.github/workflows/submodules.yml#L28)
[composer update action](https://github.com/matomo-org/matomo/blob/4.x-dev/.github/workflows/composer-update.yml#L28)
And the CLDR action as soon as it was merged.

PR #653, opened 2022-08-17
Documenting new parameter migration process

snake14:
### Description:

This documents the new migration helper class for TagManager.
This change was made as part of the TagManager changes related to issue 334. 

### Review

### DIFF
--- a/docs/4.x/tagmanager/parameters.md
+++ b/docs/4.x/tagmanager/parameters.md
@@ -122,7 +122,7 @@ $field->transform = function ($value, Setting $setting) {
 
 ### Migrating tags to include new parameters
 
-When you add a new parameter to a tag template, all the existing tags referencing this template will be shown as changed after the next version is published. They will continue to do so until each one is manually updated in some way and another version is published. To avoid this, in Matomo version 4.12 forward, there is a helper class that can be used in migration scripts.
+When you add a new parameter to a tag template, all the existing tags referencing this template will be shown as changed after the next version is published. They will continue to do so until each one is manually updated in some way and another version is published. To avoid this, there is a helper class that can be used in the migration script (Available Matomo version 4.12 and above).
 
 * First, you need to create a migration script following the process outlined in the [Extending the database](https://developer.matomo.org/guides/extending-database) page.
 * Then, in the doUpdate method, you can use the new migration helper class to update all of the tags that need to be updated.
--- a/docs/4.x/tagmanager/parameters.md
+++ b/docs/4.x/tagmanager/parameters.md
@@ -120,6 +120,28 @@ $field->transform = function ($value, Setting $setting) {
 }
 ```
 
+### Migrating tags to include new parameters
+
+When you add a new parameter to a tag template, it can cause some strange behavior if the existing tags of the template type aren't migrated. To aid in this, there is a helper class that can be used in migration scripts.
+
+* First, you need to create a migration script following the process outlined in the [Extending the database](https://developer.matomo.org/guides/extending-database) page.
+* Then, in the doUpdate method, you can use the new migration helper class to update all of the tags that need to be updated.
+```php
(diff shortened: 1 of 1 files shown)


docs/4.x/tagmanager/parameters.md
```diff
@@ -120,6 +120,28 @@ $field->transform = function ($value, Setting $setting) {
 }
 ```
 
+### Migrating tags to include new parameters
+
+When you add a new parameter to a tag template, it can cause some strange behavior if the existing tags of the template type aren't migrated. To aid in this, there is a helper class that can be used in migration scripts.
```
AltamashShaikh: @snake14 Strange behaviour won't highlight what is the issue, its better to mention what issue it will create as the user won't have context like us
```suggestion
When you add a new parameter to a tag template, all the existing tags referencing this template will be shown as changed. To avoid this, there is a helper class that can be used in migration scripts.

File: docs/4.x/tagmanager/parameters.md [hunk 2]
```diff
@@ -120,6 +120,28 @@ $field->transform = function ($value, Setting $setting) {
...
+    $updater->executeMigrations(__FILE__, $this->getMigrations($updater));
+
+    // Migrate the Matomo type tags to all include the newly configured field.
+    $migrator = new NewTagParameterMigrator(MatomoTag::ID, 'someNewParameterName');
+    $migrator->addField('anotherNewParameterName'); // This is optional and only needed if you added more than one parameter.
+    $migrator->migrate(); // This kicks off the processing of the tag migration.
+}
+```
+* If you're updating a different template class than `MatomoTag`, simply substitute the correct class, which should have the `ID` constant.
+* Note that there's an `addField` method that can be called as many times as necessary if you added more than one new parameter. If you only added one, there's no need to use the `addField` method.
```
AltamashShaikh: Make the word `Note` bold like "**Note**"

AltamashShaikh (commented):
Good work :+1: 
Just left few comments

docs/4.x/tagmanager/parameters.md
snake14: @AltamashShaikh Thank you. That's good feedback. I've tried to incorporate your suggestion.

docs/4.x/tagmanager/parameters.md
snake14: Good idea. I went ahead and made that change.

snake14:
@AltamashShaikh Thank you for your feedback. I tried to incorporate your recommendations. I also remembered that I didn't mention that this change will be available with version 4.12 forward, so I added that.

docs/4.x/tagmanager/parameters.md
```diff
@@ -120,6 +120,28 @@ $field->transform = function ($value, Setting $setting) {
 }
 ```
 
+### Migrating tags to include new parameters
+
+When you add a new parameter to a tag template, all the existing tags referencing this template will be shown as changed after the next version is published. They will continue to do so until each one is manually updated in some way and another version is published. To avoid this, in Matomo version 4.12 forward, there is a helper class that can be used in migration scripts.
```
AltamashShaikh: @snake14 Does this look better to you ?
```suggestion
When you add a new parameter to a tag template, all the existing tags referencing this template will be shown as changed after the next version is published. They will continue to do so until each one is manually updated in some way and another version is published. To avoid this, there is a helper class that can be used in the migration script (Available from Matomo 4.12 and above versions).

AltamashShaikh (commented):
@snake14 added 1 more comment

docs/4.x/tagmanager/parameters.md
snake14: @AltamashShaikh Thank you. That does flow better than what I had. I made that change with a slight variation.

AltamashShaikh (approved):

PR #678, opened 2022-11-22
Matomo5 doc

mattmary:
https://innocraft.atlassian.net/browse/WBS-573


lance-matomo:
I noticed that the default version displayed after the update is 5.x I'm not sure if this is what is wanted?

mattmary:
@lance-matomo As far I understood, we want to upgrade smoothly for the compatibility plugins: We are not in a beta state yet, but for the beta release we say:

> Now that we have made all our plugins (including premium features) compatible with this version and have released a new version, we can release the first beta for core. This allows most people to upgrade smoothly. Users will run into problems though when they use third party plugins that aren't compatible with this new major version yet. This means features/plugins will be disabled for them, which can cause them issues. So we rather release a first beta bit later in the process in the hope that some plugins are already compatible.

So I understand switching to the next major release helps to upgrade smoother to our customers.
I'm right, @tsteur ?

tsteur:
It maybe depends how far away the first beta of Matomo 5 is. Once the first beta has been released, it should be fine to have Matomo 5 selected by default. Before that it's not ideal but could be acceptable in this case, also since there's likely almost no breaking changes and barely any docs will change. If easily doable, could change the behaviour though and select Matomo 4 by default until another date. If we decide for that then we should make sure to update the process in https://developer.matomo.org/guides/release-management#new-major-releases as it'll add an extra step of making Matomo 5 the default.

@mattmary above highlighted paragraph is more about the core/plugin development and that we need to ensure that all our plugins are compatible with Matomo 5 and released on the marketplace before releasing the first beta. Otherwise, some plugins would be disabled when they are not compatible.

mattmary:
@tsteur just wondering if we did not do this documentation upgrade too early? If you expect to switch to the new release when a beta is published (This is not the case for now), are we too early?
Or for you, it's not too early, and we have to manage this use case when beta is not published yet?
@lance-matomo can you tell me where is the reference on the 5.x release you talked about?

lance-matomo:
Hey @mattmary when I bring up my local version of this:
https://local.developer.matomo.org/
In the top right in the dropdown it shows "Matomo 5.X" as the default

tsteur:
@mattmary It's not too early, it's a good time to start adding this so we can already make changes to the 5.X docs while the teams work on it. As it be only for a short time it could be ok to show Matomo 5 by default, but no preference.

mattmary:
As we have a lot of users who developed plugins, I think this is not too early too.
@lance-matomo do you want me to revert the default version to the 4.x one anyway?

lance-matomo:
yea it might be ok to leave the 5.x as the default, it will indicate to developers that 5.x is coming soon

lance-matomo (approved):

PR #691, opened 2023-01-11
Tracking API: Add note about expected URL format

pixelbrackets:
The tracking API expects URLs without trailing slashes only. Otherwise the request is tracked, but is not displayed correctly in the behaviour > pages section.

Example issue: https://forum.matomo.org/t/shows-index-for-many-pages/45127

### DIFF
--- a/docs/5.x/tracking-api.md
+++ b/docs/5.x/tracking-api.md
@@ -19,7 +19,7 @@ _Note: all parameters values that are strings (such as 'url', 'action\_name', et
 ### Recommended parameters
 
 * `action_name` **(recommended)** &mdash; The title of the action being tracked. It is possible to [use slashes / to set one or several](https://matomo.org/faq/how-to/#faq_62) [categories for this action](https://matomo.org/faq/how-to/#faq_62). For example, **Help / Feedback** will create the Action **Feedback** in the category **Help**.
-* `url` **(recommended)** &mdash; The full URL for the current action.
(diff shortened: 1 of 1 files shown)


sgiehl:
@pixelbrackets Actually I think it is correct the way it is.
If a url like `http://example.com/folder/` is tracked, Matomo will identify that as a call to the index of `folder`, which is litterally true.
Calling `http://example.com/folder` is not the same, as it actually queries for a file named `folder` in the root directory.

Or did I misunderstood something here?

pixelbrackets:
The issue described (and what I can reproduce) is this unexpected API experience:

When you send https://example.com/folder/, https://example.com/folder/subpage/, https://example.com/folder/subpage/yet-another-page/ as `url`, then all three requests are mapped to `/index`. So all three speaking URLs are mapped to a single page.

Only when you send something like https://example.com/folder/index, https://example.com/folder/subpage/index, https://example.com/folder/subpage/yet-another-page/index, they are show as different pages in the behaviour section. Same for https://example.com/folder, https://example.com/folder/subpage, https://example.com/folder/subpage/yet-another-page.

sgiehl:
All urls should be tracked as they were sent. The hierarchical report view for behavior reports might though group them by folders. I'm not able to see any problems there.
Are you maybe able to describe your problem a bit more, maybe with some screenshots? Might make it easier to understand maybe.

pixelbrackets:
The issue seems to have vanished, I can not reproduce it anymore in the latest Matomo version. Closing this PR. ¯\_(ツ)_/¯

PR #710, opened 2023-04-14
For Matomo Tag Manager users, reading the JS Tracker reference, suggest the correct approach

mattab:


AltamashShaikh (approved):
Good to roll back this info

snake14 (approved):
With the idea that we want these developer docs to be strictly customer facing, these changes make sense. The same changes will need to be made to the 4.x documentation as well.

Does everything look alright to you @AltamashShaikh ? I updated the Jira issue created to follow-up on the MTM `_paq` changes to include adding an inline comment in the core JS tracker so that core devs know that tracker configurations need to be added to MTM as well. We also have a wiki page. It was referring to this page, but I copied the content into the wiki page.

AltamashShaikh:
@snake14 Now did you mention it, we should update the doc for `4.x-dev` branch too

snake14:
> @snake14 Now did you mention it, we should update the doc for `4.x-dev` branch too

Thanks @AltamashShaikh . If the changes look good, I can make the same changes to 4.x as well.

AltamashShaikh:
> > @snake14 Now did you mention it, we should update the doc for `4.x-dev` branch too
> 
> Thanks @AltamashShaikh . If the changes look good, I can make the same changes to 4.x as well.

Yes go ahead

PR #722, opened 2023-05-06
Correct submodule url

EPNW-Eric:
It seems that the old submodule path does not work anymore and points to the old place of the repository. I've updated it accordingly.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,3 @@
 [submodule "piwik"]
 	path = piwik
-	url = git@github.com:piwik/piwik.git
+	url = https://github.com/matomo-org/matomo.git


michalkleiner:
Hi @EPNW-Eric, thank you for opening the pull request.
Have you tried if simply changing the repo name and the organisation without changing the protocol would also work? I.e. not adding the https bit.

`url = git@github.com:matomo-org/matomo.git`

EPNW-Eric:
It didn't work for me without changing the protocol:
```cmd
git clone git@github.com:matomo-org/matomo.git
Cloning into 'matomo'...
git@github.com: Permission denied (publickey).
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
```

Could also be a mistake on my end, but using `https` it worked.

sgiehl:
using git or https only makes a difference in the way the used git client tries to authenticate. As both repos are public that actually shouldn't make that much difference, but using https should be fine as well I guess.

PR #725, opened 2023-05-12
Remove return always true from composer install

lance-matomo:
### Description:

https://innocraft.atlassian.net/browse/WBS-1472
when running ./generate.sh the composer install command wasn't completing before the generate.php line was executing so it wasn't using the updated files. Extending the sleep seems to solve the issue.

### Review

### DIFF
--- a/generate.sh
+++ b/generate.sh
@@ -29,9 +29,8 @@ function generateDocs {
     sleep 3
     git submodule update --recursive --force
     php ../app/composer.phar self-update
-    php ../app/composer.phar install || true
+    php ../app/composer.phar install
     cd ..
-    sleep 10
     php generator/generate.php --branch=$1 --targetname=$2
 
     GENERATION_SUCCESS=$?
--- a/generate.sh
+++ b/generate.sh
@@ -31,7 +31,7 @@ function generateDocs {
     php ../app/composer.phar self-update
     php ../app/composer.phar install || true
     cd ..
-    sleep 4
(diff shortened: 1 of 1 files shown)


mattmary (approved):

michalkleiner:
I think that might just be hiding the problem rather than solving it.
Do we know why we do the `|| true` in the previous `php ../app/composer.phar install || true` command? It basically says run the install but always succeed regardless of the output from composer. If we just waited for that to finish and only continued when successful that should do it without any waiting afterwards? And if the install fails we would know straight away that if failed for whatever reason.

lance-matomo:
Hey @michalkleiner , good spotting on that one. I've tested your ideal locally and it seems to work. I don't know why it was written like that initially, it does seem like a workaround for something. I'll update the PR then we can see if it works on the live server.

generate.sh
```diff
@@ -29,9 +29,8 @@ function generateDocs {
     sleep 3
     git submodule update --recursive --force
     php ../app/composer.phar self-update
-    php ../app/composer.phar install || true
+    php ../app/composer.phar install
     cd ..
-    sleep 4
```
michalkleiner, line 34: We might be able to keep the sleep here (for whatever reason it was added there).
I think the problem was that it was trying to run the generator before the install finished, so removing `|| true` above should do the trick on its own.

michalkleiner (approved):

PR #754, opened 2023-08-30
Add a note that config overrides need to be in the same section

michalkleiner:
As per the title — ref: https://github.com/matomo-org/matomo/issues/21205


matomoto:
Please put an example under the note like this:
```
[General]
multisites_refresh_after_seconds = 0
site_selector_max_sites = 50
```
The most variables in the section `[General]`.

michalkleiner:
@matomoto I think the documentation on the ini config format is quite clear already and doesn't need further additions.

PR #760, opened 2023-09-19
Fix default cookie prefix _pk_

### LINKED ISSUE #758: Default cookie prefix _pk_ not shown correctly on documentation website
The default cookie prefix is `_pk_`, the documentation website shows it as _pk_ (see [here](https://developer.matomo.org/api-reference/tracking-javascript#configuration-of-tracking-cookies)).
I guess the problem is that ' are used (whiche makes the text cursive in markdown) instead of `.
The respective line is
https://github.com/matomo-org/developer-documentation/blob/75999a432b31f0ecf445162fe3d810c2cd651f8c/docs/5.x/tracking-javascript.md?plain=1#L146

EPNW-Eric:
fixes #758.


sgiehl:
Thanks for this contribution @EPNW-Eric 🎉

PR #767, opened 2023-11-05
Add Matomo Java Tracker documentation

dheid:
Hi!

I maintain the official Matomo Java Tracker and added a reference / guide on how to use the new version 3.0.6. In addition I allowed myself to edit your README.md and the PHP Composer installation (usually installed system wide).

Would be great if you could include the changes into the documentation. Thank you very much!

Best regards

Daniel


generate.sh
```diff
@@ -28,8 +28,8 @@ function generateDocs {
     git pull origin $1
     sleep 3
     git submodule update --init --recursive --force
-    php ../app/composer.phar self-update
-    php ../app/composer.phar install
+    composer update
```
sgiehl: This is actually a major change. Before composer was only updating itself. Now it will update all used packages. Not sure if that was changed on purpose.

docs/4.x/tracking-java.md
```diff
@@ -0,0 +1,169 @@
...
+---
+# Java Tracker
+
+The Matomo Java Tracker is a Java library to send tracking data to Matomo servers. It is compatible with Java 1.8 and above. You can use it to track data from any Java application, including Java web applications, Java desktop applications, Java command line applications, Java batch applications, Java microservices and more.
+
+Due to the nature of Java applications, the API is different from the JavaScript Tracker API. Nevertheless it supports synchronous and asynchronous tracking, custom dimensions, custom variables, custom events, content tracking, ecommerce tracking, goal tracking, site search tracking, campaign tracking, and more.
+
+The primary source to look for an API documentation is the [JavaDoc](https://matomo-org.github.io/matomo-java-tracker/javadoc/index.html). This guide provides a quick overview of the most important classes and methods of the library.
+
+If you experience any problems or have any questions, please [create an issue](https://github.com/matomo-org/matomo-java-tracker/issues/new) in the [Matomo Java Tracker repository]. We are happy to help you.
```
sgiehl: Should `[Matomo Java Tracker repository]` be linked? Or should the brackets be removed again?

docs/4.x/tracking-java.md
```diff
@@ -0,0 +1,169 @@
...
+
+Please mind that the token should be transmitted securely, for example using HTTPS. 
+
+## Custom Tracking Parameters
+
+You can also set additional custom request parameters using the method `MatomoRequest.request().additionalParameters(Map<String, Collection<Object>>)`. The additional parameters will be added to the request URL as query parameters. The query parameter name will be the key of the map entry and the query parameter value will be the value of the map entry. To add the same query parameter name with multiple values, you can use a collection as the parameter value, e.g. `additionalParameters(Map.of("foo", List.of("bar", "baz")))` will add the query parameters `?foo=bar&foo=baz` to the request URL (note that the order of the query parameters is not guaranteed). Note: If you set any parameter that is not supported by the Matomo tracking API, the parameter will be ignored by Matomo, although it will be added to the request URL.
```
sgiehl, line 154: > `additionalParameters(Map.of("foo", List.of("bar", "baz")))` will add the query parameters `?foo=bar&foo=baz`

What's the purpose of setting the same parameter multiple times? Wouldn't always the last value be used in that case?
Or should that be something like `?foo[]=bar&foo[]=baz` to allow sending array values?

docs/4.x/tracking-java.md
```diff
@@ -0,0 +1,169 @@
...
+
+## Ecommerce Items
+
+Matomo supports ecommerce tracking, i.e. tracking of ecommerce actions like adding items to the cart, removing items from the cart, and purchasing items. Every ecommerce action can be tracked as a goal. To track ecommerce actions, you need to set a goal id in the request. You can set the goal id using the method `MatomoRequest.request().goalId(int)`. The goal id is the id of the goal in Matomo. You can find the goal id in the Matomo user interface under "Administration" -> "Websites" -> "Goals". The goal id is the number in the URL of the edit page.
+
+Ecommerce items can be added to a request using the method `MatomoRequest.request().ecommerceItems(EcommerceItems)`. The items will be added to the request URL as a JSON representation in the query parameter `ec_items`. The JSON representation is an array of JSON objects. Each JSON object represents an ecommerce item. Ecommerce items are only supported if a goal id is set in the request. If you set ecommerce items without a goal id, the items will be ignored by Matomo.
```
sgiehl, line 162: I actually don't understand the reason of requiring a goal id. For ecommerce actions the goal id `0` needs to be sent. Not sure what sending a goal id with another value might cause. So why isn't the goal id automatically set when a ecommerce action is tracked?

docs/4.x/tracking-java.md
```diff
@@ -0,0 +1,169 @@
...
+
+To create a custom variables object use the constructor `new CustomVariables()` and add custom variables using the method `CustomVariables.add(String, String)`:
+
+```java
+CustomVariables customVariables = new CustomVariables()
+    .add("Type", "Customer")
+    .add("Status", "Logged In");
+```
+
+The variables get the index in the order of their addition. The first variable has the index 1, the second variable has the index 2, and so on. The index is used to identify the variable in the Matomo user interface. You can find the custom variables in the Matomo user interface under "Visitors" -> "Custom Variables". The index is the number in the URL of the custom variables page.
```
sgiehl: Just out of curiosity, how would you handle it if you only want to set the custom variable with index 1 and 3? Set 2 with a null/empty value? Would the empty variable then be sent with the request?

sgiehl (changes requested):
Had a rough look through the changes. Haven't done JAVA for quite a long time, so can't say anything about the code examples.

generate.sh
dheid: Yes, that purpose. I think it's basic to install the PHP Composer and it's mostly installed on a PHP developers system. I was quite confused to have a separate step to install it. Also the checksum does not match necessarily. Also you can't say if the newest version works every time. In the README I wrote that one should install PHP Composer.

sgiehl: Ok. But if we expect composer to be installed, shouldn't a `composer install` be enough? I wonder why generating the docs should execute a `composer update` to update all dependencies of Matomo. This could actually break things if a newer version of a dependency might cause problems.

docs/4.x/tracking-java.md
dheid: I agree, that it makes no sense. It's just the old behaviour of the tracker and we wanted the clients to be compatible.

docs/4.x/tracking-java.md
dheid: That's a nice idea! I opened an issue here: https://github.com/matomo-org/matomo-java-tracker/issues/162

docs/4.x/tracking-java.md
dheid: Oh, yes that an old mistake of mine. I already changed that in the Matomo Java Tracker. I will update the documentation.

docs/4.x/tracking-java.md
sgiehl: In that case it's "fine". But maybe you should consider changing that in the java tracker with the next major release or so...

generate.sh
dheid: The composer update was necessary on my machine, because composer.lock was out of sync. It cannot harm in any case.

docs/4.x/tracking-java.md
dheid: Fixed that.

docs/4.x/tracking-java.md
dheid: I opened a ticket: https://github.com/matomo-org/matomo-java-tracker/issues/163

docs/4.x/tracking-java.md
dheid: Done

dheid:
@sgiehl Thanks for looking at this PR. I updated it. Is it now mergeable?

generate.sh
sgiehl: We have the `composer.lock` contained in our repo on purpose, so we can ensure a `composer install` will always install the same versions. This was done to ensure that the versions are fully working with out code. For this reason it would be better not to do a `composer update` at that point.

docs/4.x/tracking-java.md
```diff
@@ -0,0 +1,169 @@
...
+
+To create a custom variables object use the constructor `new CustomVariables()` and add custom variables using the method `CustomVariables.add(String, String)`:
+
+```java
+CustomVariables customVariables = new CustomVariables()
+    .add("Type", "Customer")
+    .add("Status", "Logged In");
+```
+
+The first parameter contains the custom variable consisting of the key and the value. The second parameter contains the index of the custom variable. The index is used to identify the variable in the Matomo user interface. You can find the custom variables in the Matomo user interface under "Visitors" -> "Custom Variables". The index is the number in the URL of the custom variables page.
```
sgiehl, line 202: Isn't the example above now incorrect?

dheid: Argh. I only updated version 5... 😖 Will update other versions as well

dheid: Done. Something else or can we merge?

generate.sh
dheid: Removed the composer update. Anything else?

sgiehl (commented):
Documentation wise this looks good to me now. Not sure if the changes to the generation script could cause any problems on our server running it. Maybe @tsteur knows that...

dheid:
Okay, I'll remove those changes.

dheid:
And I would like to add a recent feature to the documentation if it's okay?

sgiehl:
Feel free to add something. Btw. it might not be needed to apply further changes to `2.x` and `3.x`, as those versions are not maintained any longer.

dheid:
@sgiehl Now it should be okay, please?

sgiehl (approved):

dheid:
Thanks so much!!! ❤️

PR #790, opened 2024-05-21
Update generate and push script

Frankie75321:
### Description:

Fixes buggy script and turns off errors so they won't show in between updating the server and upgrading composer.

### Review

### DIFF
--- a/generateAndPush.sh
+++ b/generateAndPush.sh
@@ -2,10 +2,6 @@
 
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 cd $DIR
-cd piwik
-git checkout live
-git pull origin live
-cd ..
 git checkout docs/*/generated
 git pull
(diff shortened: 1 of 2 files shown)


generateAndPush.sh
```diff
@@ -2,10 +2,6 @@
 
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 cd $DIR
-cd piwik
-git checkout live
```
AltamashShaikh, line 6: @Frankie75321 Rather than deleting this, should we change the branch to `5.x-dev` ?

Frankie75321: That's exactly what I suggested to Thomas but his reply was "I would probably remove those 4 lines as they won’t work anyway and shouldn’t be needed"

AltamashShaikh: Okay :+1:

Issue #793, opened 2024-07-11
Misleading documentation when it comes to custom dimension at the action level along with goals


Chardonneaur:
Hi team,

The following documentation:
![Screenshot from 2024-07-11 18-18-09](https://github.com/matomo-org/developer-documentation/assets/28002238/d52ce689-65c9-4f1e-98f1-169b84471473)
is misleading because for goals, it does not work at the custom dimension level. Only at the visit level.
Source: https://developer.matomo.org/guides/tracking-javascript-guide

Suggested solution:
To precise that it works only for the visit scope level.

Issue #791, opened 2024-07-04
Goal missing parameters


Chardonneaur:
Hi Team,

If one click within the goals report and export one of the report filtering goals:
![Screenshot from 2024-07-04 12-11-59](https://github.com/matomo-org/developer-documentation/assets/28002238/a5f2bfd1-1c17-49b7-8310-2479b13602fe)
We can notice two parameters:
filter_update_columns_when_show_all_goals=
filter_show_goal_columns_process_goals=
which are not listed within https://developer.matomo.org/api-reference/reporting-api

Could you please document them?

PR #801, opened 2024-08-26
Gill developer docs

gillnd:
### Description:

Gillian Douglas made 3 changes

### Review

1. Add goal parameter description to piwiks-reporting-api.md (KB-869, github #791 )
2. Add format_metrics and description to reporting-api.md (KB-873, github #795 )
3. Add note on visit scope level under "Tracking a custom dimension for one specific action only" (KB-870, github #793 )

### DIFF
--- a/docs/5.x/reporting-api.md
+++ b/docs/5.x/reporting-api.md
@@ -56,7 +56,7 @@ _referrerName==Google,referrerName==Bing;country==IN_
         *   see [segmentation documentation](/api-reference/segmentation) for the list of available dimensions & metrics, example values for each, and more information about the custom segment parameter
 
 
-*   **format**; defines the format of the output
+*   **format**; defines the format of the output.
 
     *   xml
     *   json (if you want to do [cross domain request in ajax](https://bob.pythonmac.org/archives/2005/12/05/remote-json-jsonp/) and get json data, you can wrap the json data around a function call by using the **jsoncallback** parameter)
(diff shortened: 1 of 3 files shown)


snake14 (approved):
Changes look good. Tested locally and everything appears to display correctly :+1:

docs/5.x/reporting-api.md
```diff
@@ -65,7 +65,9 @@ _referrerName==Google,referrerName==Bing;country==IN_
     *   html
     *   rss (when **date** is a range for example date=last10 or date=previous15)
     *   original; to fetch the original PHP data structure. This is useful when you call the Piwik API [internally using the PHP code](/guides/querying-the-reporting-api)
-    
+
+*   **format_metrics**; defines the output format of the metrics' values.
+    *   By appending `&format_metrics=0`, the metrics' values are displayed as numerical values in the output instead of the default percent rate.
```
sgiehl, line 70: Just for completeness: There are actually three possible values that can be provided here:
* `format_metrics=0`: metrics will not be formatted at all. Percent values are returned as rates - This might be best used if the numbers should be used for further processing
* `format_metrics=1`: All metrics are returned formatted (this also includes appending percent or money signs) - This might be best used if the data should be used for direct presentation
* `format_metrics=bc`: In that case only percent values are formatted, while all other values aren't. This behavior is deprecated and only still exists, as the Matomo UI expects it in some places. This is also the default behavior if nothing is provided. At some point this option will be removed and `format_metrics=0` will become the new default.

gillnd: @sgiehl Thank you. I will add this

snake14: Apologies @gillnd . I guess I should have waited to merge your PR.

gillnd: @snake14 no worries at all, thanks for all your assistance. I've added sgiehl's comments in a [new pull request 801](https://github.com/matomo-org/developer-documentation/pull/802)

PR #802, opened 2024-08-26
Include additional values and description for format_metrics in repor…

gillnd:
Updates to reporting-api.md

Updated with comment from sgiehl
1) `format_metrics=0`: metrics will not be formatted at all. Percent values are returned as rates. This can be applied if the numbers should be used for further processing. 
2) `format_metrics=1`: All metrics returned are formatted (this also includes appending percent or money signs). This can be applied if the data should be used for direct presentation. 
3) `format_metrics=bc`: In this case, only percent values are formatted, while all other values are not. This behaviour is deprecated but still exists as the Matomo UI requires it in specific instances. It is the default behaviour if nothing is provided. In the future, this option will be removed and `format_metrics=0` will become the new default.


snake14 (approved):
Looks good :+1:

snake14:
@sgiehl I'll hold off on merging until you've had a chance to confirm it has all that is needed.

sgiehl (approved):
Looks good to me.

PR #829, opened 2025-07-22
WBS-813: Fix for duplicate anchor IDs

textagroup:
### Description:

Some IDs for anchor links are being duplicated.
A good example of this is the **Breaking Changes** links on the changelog.
This is causing a few W3 errors https://validator.w3.org/nu/?doc=https%3A%2F%2Fdeveloper.matomo.org%2Fchangelog

This should also fix the issue of the anchor links not working.
There are a few instances where anchors may not work but this may be related to errors in how the Markdown file was created,
an example of this is some of the links for Piwik 2.16.1 which has been duplicated.

image

### Review

### DIFF
--- a/app/helpers/Content/Guide.php
+++ b/app/helpers/Content/Guide.php
@@ -225,7 +225,7 @@ class Guide implements MenuItem
     {
         $anchorList = [];
         foreach ($anchors as $anchor) {
-            $anchorList[$anchor['sub']][] = $anchor['parent'];
+            $anchorList[$anchor['subsectionId']][] = $anchor['sectionId'];
         }
         $dom = new \DomDocument();
         // Will probably be duplicate ID warnings so suppress them
--- a/app/helpers/Content/Guide.php
+++ b/app/helpers/Content/Guide.php
@@ -43,6 +43,16 @@ class Guide implements MenuItem
     /**
      * @return string
      */
+    public function getValidatedRenderedContent($anchors = [])
+    {
+        $content = $this->getRenderedcontent();
+        $content = self::fixDuplicateAnchorIds($content, $anchors);
+        return $content;
+    }
+
+    /**
+     * @return string
+     */
     public function getTitle()
     {
         if (isset($this->document->metadata['title'])) {
@@ -206,5 +216,30 @@ class Guide implements MenuItem
 
             return $this->getTitle();
         }
+    }
+
+    /**
+     * @return string
+     */
+    public static function fixDuplicateAnchorIDs($content, $anchors)
+    {
+        $anchorList = [];
+        foreach ($anchors as $anchor) {
+            $anchorList[$anchor['sub']][] = $anchor['parent'];
+        }
+        $dom = new \DomDocument();
+        @$dom->loadHtml($content);
+        $anchors = $dom->getElementsByTagName('*');
+        $ids=[];
+        foreach ($anchors as $anchor) {
+            $id = $anchor->getAttribute('id');
+            if ($id && isset($anchorList[$id]) && in_array($id, $ids)) {
+                $parent = array_shift($anchorList[$id]);
+                $anchor->setAttribute('id', "$parent-$id");
+            }
+            $ids[] = $id;
+        }
+
+        return $dom->saveHTML();
     }
 }
--- a/app/helpers/Content/Guide.php
+++ b/app/helpers/Content/Guide.php
@@ -228,14 +228,23 @@ class Guide implements MenuItem
             $anchorList[$anchor['sub']][] = $anchor['parent'];
         }
         $dom = new \DomDocument();
+        // Will probably be duplicate ID warnings so suppress them
         @$dom->loadHtml($content);
         $anchors = $dom->getElementsByTagName('*');
         $ids=[];
+        $duplicateId = 0;
         foreach ($anchors as $anchor) {
             $id = $anchor->getAttribute('id');
             if ($id && isset($anchorList[$id]) && in_array($id, $ids)) {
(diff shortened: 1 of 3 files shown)


textagroup:
Another possible way to tackle this
https://github.com/michelf/php-markdown/issues/310

Frankie75321 (changes requested):
The menu links are all correct but some of the ID's in the content are not right.
For example, if you click on Matomo 3.10.0 -> New APIs, you get taken to the Matomo 5.0.0
 section.

textagroup:
> The menu links are all correct but some of the ID's in the content are not right. For example, if you click on Matomo 3.10.0 -> New APIs, you get taken to the Matomo 5.0.0 section.

I don't think this is related to the pull request but is being caused by issues with the Markdown the HTML is being generated from.
A good example of this is for **Matomo 5.1.0** which has the following links in the left menu

- Breaking Changes
- Deprecations
- Javascript Tracker

But when you actually go to that section it also has a section **New APIs** so you have 27 links in the left hand menu but there are actually 32 possible anchors.

I suspect there will be quite a few errors like this in the Markdown.
@michalkleiner We may need to perform some extra checks on the changelog PRs to try and improve the HTML that gets generated for the example I have given (5.10) the issue is being caused by a extra # so instead of a h3 a h4 is used.


```
### Deprecations

The API method `Overlay.getExcludedQueryParameters` has been deprecated and will be removed in Matomo 6. Use the new method `SitesManager.getExcludedQueryParameters` instead.

### JavaScript Tracker

#### New APIs

* The method `disableCampaignParameters` have been added to the JavaScript tracker. It allows to disable processing of campaign parameters and forwarding them to the tracking endpoint.`


https://github.com/user-attachments/assets/edb92857-9ed2-446e-8292-9d0663ca8996
```


https://github.com/user-attachments/assets/8e296867-2dca-4dc7-aec1-6ad63051ec8d

textagroup:
> The menu links are all correct but some of the ID's in the content are not right. For example, if you click on Matomo 3.10.0 -> New APIs, you get taken to the Matomo 5.0.0 section.

Thinking about this last night, I can probably check that the element is a h3 before amending the ID.
If it is a h4 might be an idea to just tweak the ID so it is at least unique

app/helpers/Content/Guide.php
```diff
@@ -207,4 +217,29 @@ public function getSEOTitle()
...
     }
+
+    /**
+     * @return string
+     */
+    public static function fixDuplicateAnchorIDs($content, $anchors)
+    {
+        $anchorList = [];
+        foreach ($anchors as $anchor) {
+            $anchorList[$anchor['sub']][] = $anchor['parent'];
```
michalkleiner: Can a subsection have multiple parents? Basically wondering if we need to store an array of parents or just a single parent value.

app/templates/guide.twig
```diff
@@ -18,7 +19,22 @@
...
             {% endif %}
 
-            {{ guide.renderedContent|raw }}
+            {% if guide.sections|length %}
+                {% set links = [] %}
+                {% for section in guide.sections %}
+                    {% if section.children|length %}
+                        {% for subsection in section.children %}
+                            {% if subsection.id in links %}
+                                {% set anchorLinks = anchorLinks|merge([{ parent: section.id, sub: subsection.id,  }]) %}
```
michalkleiner: I'd be more descriptive with the array keys and call them `parentId` (or `sectionId`) and `subsectionId`.

app/helpers/Content/Guide.php
textagroup: A subsection will only have one parent

app/templates/guide.twig
textagroup: Updated

app/helpers/Content/Guide.php
```diff
@@ -207,4 +217,38 @@ public function getSEOTitle()
...
+                    $parent = array_shift($anchorList[$id]);
+                    $anchor->setAttribute('id', "$parent-$id");
+                } else {
+                    // if it is not a h3 amend so the ID is not duplicated
+                    $duplicateId++;
+                    $anchor->setAttribute('id', "$id-$duplicateId");
+                }
+            }
+            $ids[] = $id;
+        }
```
michalkleiner, line 250: a suggestion to not store the parent in an array

```suggestion
        foreach ($anchors as $anchor) {
            $anchorList[$anchor['subsectionId']] = $anchor['sectionId'];
        }
        $dom = new \DomDocument();
        // Will probably be duplicate ID warnings so suppress them
        @$dom->loadHtml($content);
        $anchors = $dom->getElementsByTagName('*');
        $ids = [];
        $duplicateId = 0;
        foreach ($anchors as $anchor) {
            $id = $anchor->getAttribute('id');
            if ($id && isset($anchorList[$id]) && in_array($id, $ids)) {
                // should only amend the ID via $anchorList if it is a h3
                if ($anchor->tagName == 'h3') {
                    $parentId = $anchorList[$id];
                    $anchor->setAttribute('id', "$parentId-$id");
                } else {
                    // if it is not a h3, amend so the ID is not duplicated
                    $duplicateId++;
                    $anchor->setAttribute('id', "$id-$duplicateId");
                }
            }
            $ids[] = $id;
        }
```

textagroup: Tested it and it does not work the anchors end up all getting the same id

michalkleiner: All good, thanks for giving it a go. I don't have an instance up to debug this further, so let's leave it as is.

Frankie75321 (approved):
Goods good locally.

PR #842, opened 2025-09-26
Fixed issue with local environment caching routes

snake14:
### Description:

While developing some new routes, I noticed that despite having `CACHING_ENABLED` set to false in my local config, I had to manually clear the route cache file each time I made a route path change. The line I changed appears to be doing the opposite of what you would expect and only caching routes when caching is disabled.

### Review

### DIFF
--- a/app/public/index.php
+++ b/app/public/index.php
@@ -69,7 +69,7 @@ $app->add(new MatomoVersionMiddleware());
 $app->add(new CacheMiddleware());
 
 $routeCollector = $app->getRouteCollector();
-if (!CACHING_ENABLED) {
+if (CACHING_ENABLED) {
     $routeCollector->setCacheFile('../tmp/cache/route_cache.php');
 }
 


snake14:
I tagged Core Team for review, but I just realised that maybe I should have tagged Websites Team instead. @textagroup Do you know whether there's a specific reason the `CACHING_ENABLED` constant was set the way it was? To me, it seems like it's doing the opposite of what it should be doing on the line that I'm changing in this PR.

snake14:
> I tagged Core Team for review, but I just realised that maybe I should have tagged Websites Team instead. @textagroup Do you know whether there's a specific reason the `CACHING_ENABLED` constant was set the way it was? To me, it seems like it's doing the opposite of what it should be doing on the line that I'm changing in this PR.

Any idea about my question above @Frankie75321 ?

michalkleiner:
Since @tsteur came across the same and arrived at the same solution, I think this tweak is the correct solution. If there's no objection from @textagroup or @Frankie75321 or @sgiehl, I think we can merge this. On the face of it, it makes sense without the negation.

Frankie75321:
> Since @tsteur came across the same and arrived at the same solution, I think this tweak is the correct solution. If there's no objection from @textagroup or @Frankie75321 or @sgiehl, I think we can merge this. On the face of it, it makes sense without the negation.

No objection from me. I came across the same thing.

james-hill-matomo (approved):

snake14:
Thank you for everyone's feedback. Even with the PR being approved, I don't appear to have permission to merge. @Frankie75321 or someone else with merge permissions for this repo, could you please merge this PR?

PR #876, opened 2026-03-31
Switch to proxy URL to handle API requests to demo, #PG-5029

lachiebol:
### Description

Please include a description of this change and which issue it fixes. If no issue exists yet please include context and what problem it solves.

### Checklist

### Review

### DIFF
--- a/app/routes/page.php
+++ b/app/routes/page.php
@@ -267,7 +267,7 @@ $app->get('/data/documents', function (Request $request, Response $response, $ar
         ->withStatus(200);
 });
 
-$app->get('/demo-proxy/{path:.*}', function (Request $request, Response $response, $args) {
+$app->get('/demo/{path:.*}', function (Request $request, Response $response, $args) {
     $path = $args['path'] ?? '';
 
     try {
--- a/app/routes/page.php
+++ b/app/routes/page.php
@@ -20,6 +20,7 @@ use helpers\Content\Category\SupportCategory;
 use helpers\Content\Category\TracTicketArchiveCategory;
 use helpers\Content\Guide;
 use helpers\Content\PhpDoc;
+use helpers\DemoProxy;
 use helpers\DocumentNotExistException;
 use helpers\Environment;
 use helpers\OpenApiSpecRegistry;
@@ -29,6 +30,7 @@ use helpers\Url;
 use Psr\Http\Message\ResponseInterface as Response;
 use Psr\Http\Message\ServerRequestInterface as Request;
 use Slim\Exception\HttpNotFoundException;
+
 
 
 function renderGuide(Slim\Views\Twig $view, Response $response, Psr\Http\Message\UriInterface $uri, Guide $guide, Category $category, string $template = 'guide.twig')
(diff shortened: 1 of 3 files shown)


app/templates/api-swagger.twig
```diff
@@ -19,7 +19,8 @@
     <script src="/vendor/swagger-ui/swagger-ui-bundle.js?{{ revision|e('html_attr') }}"></script>
     <script type="text/javascript">
     window.onload = function () {
-        let pluginSpec = {{ pluginSpecJson|raw }};
+        let pluginSpecJson = {{ pluginSpecJson|raw }};
+        let proxyBaseUrl = window.location.origin + '/demo-proxy';
```
AltamashShaikh:
```suggestion
        let proxyBaseUrl = '/demo-proxy';
```

app/routes/page.php
```diff
@@ -265,6 +267,20 @@ function renderGuide(Slim\Views\Twig $view, Response $response, Psr\Http\Message
...
+    $path = ltrim($args['path'] ?? '', '/');
+    $targetUrl = rtrim(DemoProxy::MATOMO_SWAGGER_PROXY_TARGET, '/') . '/' . $path;
+    $query = $request->getUri()->getQuery();
+    if ($query !== '') {
+        $targetUrl .= '?' . $query;
+    }
+
+    $proxiedResponse = DemoProxy::get($targetUrl, $request->getHeaderLine('Authorization'));
+
+    $response->getBody()->write($proxiedResponse['body']);
```
AltamashShaikh, line 287: we should add a try/catch here

app/routes/page.php
```diff
@@ -265,6 +267,20 @@ function renderGuide(Slim\Views\Twig $view, Response $response, Psr\Http\Message
         ->withStatus(200);
 });
 
+$app->get('/demo-proxy/{path:.*}', function (Request $request, Response $response, $args) {
+    $path = ltrim($args['path'] ?? '', '/');
```
AltamashShaikh: main concern is that it now becomes a proxy to query `demo.matomo.cloud` without any CORS error

We can add a check that requestURL path should be index.php and Module should be API along with CSRF token.

AltamashShaikh (commented):
Is this correct ?

Image

AltamashShaikh (changes requested):
left few comments

lachiebol:
> Is this correct ?
> 
> Image



@AltamashShaikh That's the new proxy endpoint that swagger will make requests against, do you think it should still show demo.matomo.cloud? I might check if I can change the text but not the URL

app/routes/page.php
lachiebol: Done, I also realised we return errors with a 200 OK? At the moment it doesn't break anything in Swagger, technically we could catch in the proxy and throw something else

app/routes/page.php
lachiebol: Have added some URL validation so only the API can be queried.

Demo is a public instance, are we exposing anything that isn't already available with this proxy? For the CSRF token, we don't have a session on developer docs, so not sure where we can store that

app/templates/api-swagger.twig
lachiebol: Made that change, it also collapses the server field down to this
image

I changed it to 'demo' rather than 'demo-proxy', proxy felt too technical for a user facing page.

AltamashShaikh (approved):
@lachiebol Good to add some tests with the help of AI

lachiebol:
@AltamashShaikh Sorry, just got around to this, but we don't have PHPUnit setup at all, and no testing workflow for this. Do we want to introduce testing pipelines for developer docs?

AltamashShaikh:
> @AltamashShaikh Sorry, just got around to this, but we don't have PHPUnit setup at all, and no testing workflow for this. Do we want to introduce testing pipelines for developer docs?

@lachiebol Create a follow-up ticket, we should add it at the end, nothing is needed now.

PR #949, opened 2026-04-13
Added simple script to fetch API specs, #PG-5043

lachiebol:
### Description

NOTE: This OpenApiDocs change is needed to run this script https://github.com/matomo-org/plugin-OpenApiDocs/pull/30

Script to pull specs from cloud, default MATOMO_BASE_URL will be demo.matomo.cloud

`MATOMO_BASE_URL="{matomo-url}" php fetch-openapi-specs.php`

### Checklist

### Review

### DIFF
--- a/fetch-openapi-specs.php
+++ b/fetch-openapi-specs.php
@@ -1,0 +1,81 @@
+<?php
+
+declare(strict_types=1);
+
+$baseUrl = rtrim(getenv('MATOMO_BASE_URL') ?: 'https://demo.matomo.cloud/', '/');
+$tokenAuth = getenv('MATOMO_TOKEN_AUTH') ?: 'anonymous';
+$targetDirectory = __DIR__ . '/app/public/openapi';
+$version = '1.0.0';
+
+if (!is_dir($targetDirectory)) {
+    fwrite(STDERR, "Target directory does not exist: $targetDirectory\n");
+    exit(1);
+}
+
+function fetchJson(string $baseUrl, string $tokenAuth, array $params): array
+{
+    $params['module'] = 'API';
+    $params['format'] = 'JSON';
+    $params['token_auth'] = $tokenAuth;
+
+    $url = $baseUrl . '/index.php?' . http_build_query($params);
+    $response = @file_get_contents($url);
+
+    if ($response === false) {
+        throw new RuntimeException("Request failed: $url");
+    }
+
+    $decoded = json_decode($response, true);
+    if (!is_array($decoded)) {
+        throw new RuntimeException("Invalid JSON response: $url");
+    }
+
+    if (isset($decoded['result']) && $decoded['result'] === 'error') {
+        $message = is_string($decoded['message'] ?? null) ? $decoded['message'] : 'Unknown API error';
+        throw new RuntimeException($message);
+    }
+
+    return $decoded;
+}
+
+try {
+    $plugins = fetchJson($baseUrl, $tokenAuth, [
+        'method' => 'OpenApiDocs.getPluginWhitelist',
+    ]);
+} catch (Throwable $e) {
+    fwrite(STDERR, "Failed to fetch plugin whitelist: {$e->getMessage()}\n");
+    exit(1);
+}
+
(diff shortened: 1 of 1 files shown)


fetch-openapi-specs.php
```diff
@@ -0,0 +1,81 @@
...
+}
+
+function fetchJson(string $baseUrl, string $tokenAuth, array $params): array
+{
+    $params['module'] = 'API';
+    $params['format'] = 'JSON';
+    $params['token_auth'] = $tokenAuth;
+
+    $url = $baseUrl . '/index.php?' . http_build_query($params);
+    $response = @file_get_contents($url);
```
AltamashShaikh: we should use curl ideally for remote URLs

fetch-openapi-specs.php
```diff
@@ -0,0 +1,81 @@
...
+
+        $json .= "\n";
+
+        if (file_put_contents($path, $json) === false) {
+            throw new RuntimeException("Failed to write file: $path");
+        }
+
+        $written++;
+    } catch (Throwable $e) {
+        fwrite(STDERR, "Failed for $plugin: {$e->getMessage()}\n");
```
AltamashShaikh, line 156: Here it will keep continuing unnecessarily if file_put_contents fail, the issue would be consistent for all

AltamashShaikh (changes requested):
@lachiebol left few comments

lachiebol:
@AltamashShaikh 

From codex:

```
  1. High: the fetch is only atomic per file, not for the spec set, so a mid-run failure leaves the docs site serving a mixed snapshot. In fetch-openapi-specs.php:77, each plugin
     file is replaced immediately, and on the first later failure the script exits at fetch-openapi-specs.php:111. That means earlier plugins are now from the new Matomo instance/
     version, while untouched files still represent the previous run. Since the app reads whatever is present in app/public/openapi via app/helpers/OpenApiSpecRegistry.php:21, users
     can get a partially updated API reference without any indication it is inconsistent.
```

This doesn't seem too bad for our use case right? I know we could write to a tmp folder then move after everything is done. 

Not sure if I need to address that one.

AltamashShaikh (commented):
@lachiebol AI review

```
High-risk issues (blocking)

  - Secret leakage in error output: fetch-openapi-specs.php:26 and fetch-openapi-specs.php:43 include the full request URL in thrown exceptions. That URL contains token_auth, so any network/API failure will print
    credentials to stderr via the top-level handlers at fetch-openapi-specs.php:71 and fetch-openapi-specs.php:112. For a script likely to run in CI or shared shells, that is a real credential exposure.
  - Stale spec files are never removed: the script only writes files for the current whitelist, but the docs app enumerates all matching files on disk via app/helpers/OpenApiSpecRegistry.php:21. If a plugin is removed
    from the whitelist or renamed, its old *_openapi_spec_v1.0.0.json file remains and will still be shown in the UI. That makes the generated docs drift from the source instance.
```

lachiebol:
> @lachiebol AI review
> 
> ```
> High-risk issues (blocking)
> 
>   - Secret leakage in error output: fetch-openapi-specs.php:26 and fetch-openapi-specs.php:43 include the full request URL in thrown exceptions. That URL contains token_auth, so any network/API failure will print
>     credentials to stderr via the top-level handlers at fetch-openapi-specs.php:71 and fetch-openapi-specs.php:112. For a script likely to run in CI or shared shells, that is a real credential exposure.
>   - Stale spec files are never removed: the script only writes files for the current whitelist, but the docs app enumerates all matching files on disk via app/helpers/OpenApiSpecRegistry.php:21. If a plugin is removed
>     from the whitelist or renamed, its old *_openapi_spec_v1.0.0.json file remains and will still be shown in the UI. That makes the generated docs drift from the source instance.
> ```

Secret leakage in error output: Should only be leaking token_auth anonymous but can remove that

Stale spec files are never removed: Can fix that one

AltamashShaikh (approved):
@lachiebol Looks good we can test it together after deployment and I should be able to create a cron for you

PR #967, opened 2026-07-20
Ver3 4docupdate

gillnd:
### Description
1. Updated v3 guide with the new description for Improve visit duration accuracy with the Heartbeat Timer (as per v5 text).
2. Updated v4 guide with the new description for Improve visit duration accuracy with the Heartbeat Timer (as per v5 text).
3. In v5 guide, corrected line break.

### Checklist

- [✔/✖/NA] I have understood, reviewed, and tested all AI outputs before use
- [✔/✖/NA] All AI instructions respect security, IP, and privacy rules

### Review

### DIFF
--- a/docs/3.x/tracking-javascript-guide.md
+++ b/docs/3.x/tracking-javascript-guide.md
@@ -141,7 +141,7 @@ By default, Matomo calculates visit duration from the time between tracked actio
 * A visit containing only one page view has a visit duration of 0 seconds.
 * The final page in a visit has no recorded time on page because there is no subsequent tracked action from which Matomo can calculate its duration.
 
-You can enable the Heartbeat Timer to improve visit duration tracking by sending a heartbeat ping when the visitor leaves or hides the page.
+You can enable the Heartbeat Timer to improve visit duration tracking by sending a heartbeat ping when the page loses focus or is unloaded.
 
 **Enable the Heartbeat Timer**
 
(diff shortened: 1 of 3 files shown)


caddoo (changes requested):
Superseded — see inline comments below.

docs/3.x/tracking-javascript-guide.md
```diff
@@ -134,32 +134,66 @@ _paq.push(['trackGoal', 1, <?php echo $cart->getCartValue(); ?>]);
...
+
+* The tab has remained active and in focus for at least 15 seconds.
+* The visitor then switches tabs, hides the page, navigates away, or closes it.
+* At least 15 seconds have passed since the previous tracking request.
+
+**How the Heartbeat Timer works**
+
+The Heartbeat Timer responds to browser events and does not send requests continuously at fixed intervals. A heartbeat ping is sent only when all of the following conditions are met:
+
+* the initial page view has been tracked, and
```
caddoo: Two issues in this section (3.x only):

1. **This bullet is wrong** — the tracker requires *any* prior tracking request, not specifically a page view. Change to:
   ```
   * a previous tracking request has been sent, and
   ```
   Evidence: `js/piwik.js` in 3.x (`enableHeartBeatTimer` setup uses `lastTrackerRequestTime !== null`).

2. **Remove hidden-page / minimise wording throughout the 3.x doc** (lines 144, 167, 176). The 3.x tracker only listens for `focus`/`blur`/`unload` — it does not register `visibilitychange` or inspect `document.visibilityState` (that was added in 4.x). So scenarios like "hides the page" or "minimises the browser" don't actually trigger a heartbeat in 3.x. Suggest keeping the 3.x wording to focus/blur/unload equivalents only.

docs/4.x/tracking-javascript-guide.md
```diff
@@ -147,33 +147,66 @@ _paq.push(['trackGoal', 1, <?php echo $cart->getCartValue(); ?>]);
...
+
+* The tab has remained active and in focus for at least 15 seconds.
+* The visitor then switches tabs, hides the page, navigates away, or closes it.
+* At least 15 seconds have passed since the previous tracking request.
+
+**How the Heartbeat Timer works**
+
+The Heartbeat Timer responds to browser events and does not send requests continuously at fixed intervals. A heartbeat ping is sent only when all of the following conditions are met:
+
+* the initial page view has been tracked, and
```
caddoo: This bullet is wrong for 4.x too — the tracker requires *any* prior tracking request, not specifically a page view. Change to:
```
* a previous tracking request has been sent, and
```
Evidence: `js/piwik.js` in 4.x (`enableHeartBeatTimer` setup uses `lastTrackerRequestTime !== null`).

gillnd (commented):
@caddoo Corrected as per the notes

caddoo (approved):

PR #971, opened 2026-08-24
Regenerate 6.x API reference from 6.x-dev

sgiehl:
Generated output only — no hand edits.

- `PROJECT_VERSION` now `6.x-dev`
- New: `Piwik\Http\SecurityHeaders`, `CoreHome\Columns\Metrics\PercentOfReportTotal`
- 34 files updated, 0 removed

**Blocker for future runs:** Sami crashes on PHP 8.0+ union types (`Object of class PhpParser\Node\UnionType could not be converted to string`), which Matomo 6 core uses. I patched `tsteur/sami` `Sami/Parser/NodeVisitor.php` locally to produce this output, but `generator/vendor` is gitignored, so the fix is not in this PR. Until the fork is fixed, the scheduled `generateAndPush.sh` will fail on 6.x. Two one-line `elseif` branches at the return-type (~L219) and param-type (~L175) sites.


chippison (dismissed):
There are mangled type strings in the output, but this is something already in 5.x as well and this is a long standing generator limitation. 

Nothing to be done in this PR 👍

AltamashShaikh (approved):

